PR #35, opened 2014-11-22
Don't exit on missing distribution version

vincent23:
Rolling release based distributions like Arch Linux don't have a versioning scheme and therefore don't set any version in `/etc/os-release`. Currently mason just exits if no version is set, but I think this could easily be changed into a warning.


springmeyer:
Apologies for missing commenting here sooner. I presume this is stale. Please re-open a new PR if the problem persists.

PR #45, opened 2015-01-13
Add "mason link" command

springmeyer:
This is similar to the brew link command. It sets up a local directory where all include and libs can be found.


springmeyer:
I'd like to find a way to make the bash to set up symlinks cleaner/easier to read but my only idea is to use python more (or some other language). Going to merge this now and see how it fares.

Issue #47, opened 2015-01-15
Single Linux binary


ljbade:
Related to https://github.com/mapbox/mason/issues/38

@mikemorris was running into problems where he was building everything from source since the only binaries Travis publish are for Ubuntu 12.04

Since Linux binaries are backwards compatible it makes sense to just switch to a single binary for all Linux x64 distros.

If we get feedback that some binary does not work on another distro like Fedora or such (although it should just work OK) we can revisit this.

ljbade:
I made a PR here: https://github.com/mapbox/mason/pull/48

springmeyer:
Looks like versioning ended up `linux-linux-{arch}`. Not a breaking problem, but odd looking.

ljbade:
Oh will look into it next week

ljbade:
I fixed the versioning.

PR #48, opened 2015-01-15
Use single linux binary

ljbade:
See https://github.com/mapbox/mason/issues/47

### DIFF
--- a/mason.sh
+++ b/mason.sh
@@ -71,7 +71,8 @@ elif [ ${MASON_PLATFORM} = 'linux' ]; then
         exit 1
     fi
 
-    export MASON_PLATFORM_VERSION=${MASON_PLATFORM_DISTRIBUTION}-${MASON_PLATFORM_DISTRIBUTION_VERSION}-`uname -m`
+    export MASON_PLATFORM_VERSION=linux-`uname -m`
+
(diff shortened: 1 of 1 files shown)


ljbade:
I changed the linux binary to just be `linux-arch`

@kkaefer and @springmeyer can you review?

springmeyer:
Looks good to me. As discussed on chat this change means:

  - Travis will be able to build binaries that work on Ubuntu 12.04 and greater. Therefore Ubuntu 14.04 users to start getting binaries right away. It also means that other modern distros will start working too out of the box like RHEL 7 (as long as their libc is as recent as Ubuntu 12.04, which is a very good bet).
  - When Travis upgrades to Ubuntu 14.04 this change will have the impact that we'll drop Ubuntu 12.04 support. All binaries that might get rebuilt would then be broken on Ubuntu 12.04. This is because the libc is newer on 14.04 and linux offers only forward compatibility and not back compatibility in terms of how symbol versioning works in libc.

One recommendation: I'd like Mason to keep reporting `MASON_PLATFORM_DISTRIBUTION` - seems handy to make that available even if we don't use it to version `MASON_PLATFORM_VERSION`.

ljbade:
Oh yeah Ill add that back then.

springmeyer:
:+1:

Issue #44, opened 2015-01-11
MASON_DIR refers to master only


ljbade:
I have a need to refer to files in the branch that holds the `script.sh` file. On Travis this maps to `~/build/mapbox/mason` but likely different in other contexts.

There is the variable `MASON_DIR` which maps to `~/mason.sh` which is where mason master branch is checked out.

There is no way to get the `script.sh` branch as far as I can see.

/cc @springmeyer

springmeyer:
/cc @kkaefer for ideas. My understanding is that this is by design.

> I have a need to refer to files in the branch that holds the script.sh file

I've also been keen to do this: to be able to keep patch files beside the ./script.sh but I'm not aware of a good way to do it yet.

ljbade:
all thats really needed is another variable I would think that is set to the location of the branch.

perhaps just add a `MASON_BRANCH = pwd` to the top of each `script.sh`

kkaefer:
This is by design: When installing a library with mason, we're not checking out the whole repository. Instead, we're just downloading the `script.sh` file: https://github.com/mapbox/mason/blob/861ddecd3e78ee89e2ee681509aec312810374a9/mason#L61-L65

If you need additional files, you have to download them manually in your install script.

ljbade:
@kkaefer I think @springmeyer is cooking something up in the `mason link` PR

springmeyer:
> If you need additional files, you have to download them manually in your install script.

This is what I've started doing for patch files (https://github.com/mapbox/mason/blob/gdal-1.11.1/script.sh#L32). It is a bit odd in that you need to commit the files before you can use them (making testing locally or offline impossible) but I think it should work for now.

> @kkaefer I think @springmeyer is cooking something up in the mason link PR c67cda5

@ljbade - that is unrelated to this.

> perhaps just add a MASON_BRANCH = pwd to the top of each script.sh

Yeah that might be useful.

ljbade:
hmm @springmeyer what about for libboringssl where i need to pull down a large number of different files?

ljbade:
We now pull files directly.

PR #49, opened 2015-01-16
add /lib to MASON_OPENSSL path

mikemorris:
I think this is why OpenSSL symbols were missing in `libcurl` on the `libcurl-7.38.0-boringssl` branch @ljbade, not able to test this locally though because of https://github.com/mapbox/mason/issues/44. Wonder if this is enough to fix https://github.com/mapbox/mapbox-gl-native/pull/754

### DIFF
--- a/script.sh
+++ b/script.sh
@@ -22,7 +22,7 @@ function mason_load_source {
 
 function mason_prepare_compile {
     ${MASON_DIR:-~/.mason}/mason install boringssl d3bcf13
-    MASON_OPENSSL=`~/.mason/mason prefix boringssl d3bcf13`
+    MASON_OPENSSL=`~/.mason/mason prefix boringssl d3bcf13`/lib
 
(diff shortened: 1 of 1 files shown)


springmeyer:
Does not look right to me. Asking curl about that `-with-ssl` flag gives:

```
curl-7.36.0[master]$ ./configure --help | grep ssl
  --with-ssl=PATH         Where to look for OpenSSL, PATH points to the SSL
                          installation (default: /usr/local/ssl); when
```

So that does not look to me like it is asking for the `/lib` directory but rather the directory containing `./lib` and `/includes` for openssl.

I'd recommend trying to figure out the reason linking was not working rather than guessing on a fix. The reason a configure step fails (for autotool based apps like curl) is evident in the `config.log`. When I last spoke to @ljbade the cause was clear why linking was not working: `config.log` shows that `libssl.so` was an invalid archive and so it WAS being found but was busted for some reason.

PR #64, opened 2015-01-31
Don't assume bash is at /bin/bash.

bentley:
(For example, on OpenBSD it is at /usr/local/bin/bash.)


springmeyer:
Thanks, been trying to stick with /usr/bin/env bash, but these slipped.

PR #66, opened 2015-02-03
Add tests and more android arch to zlib-system

springmeyer:
- Adds variety of tests for the zlib package.
 - Fixes reporting of paths on android for arch's that use `lib64` (like mips64)


springmeyer:
@ljbade - can I have your eyes on these android archs? Are these what we want to be testing now?

ljbade:
Yep, eventually I will get `x86-64` working too. It is held up my OpenSSL.

PR #84, opened 2015-05-13
[not ready] Install deps from config file

lbud:
@springmeyer — here's a really rough first pass at this, but do you see any major red flags here?

### DIFF
--- a/mason
+++ b/mason
@@ -41,28 +41,56 @@ elif [ "${MASON_COMMAND}" = "env" ]; then
     exit
 fi
 
-if [ -z "${MASON_NAME}" ]; then
-    usage
-    echo "Missing <lib>"
-    exit 1
-elif [ -z "${MASON_VERSION}" ]; then
-    usage
-    echo "Missing <version>"
-    exit 1
+if [ "${MASON_COMMAND}" != "install" ]; then
+    if [[ -z "${MASON_NAME}" && "${MASON_COMMAND}" != "install" ]]; then
+        usage
+        echo "Missing <lib>"
+        exit 1
+    elif [[ -z "${MASON_VERSION}" && "${MASON_COMMAND}" != "install" ]]; then
+        usage
+        echo "Missing <version>"
+        exit 1
+    fi
+elif [[ -z "${MASON_VERSION}" && -z "${MASON_NAME}" ]]; then
+    # this is install, with zero args. err if no config.json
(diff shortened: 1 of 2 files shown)


lbud:
Aha! Ok, just tested this in the wild locally (https://github.com/Project-OSRM/node-osrm/pull/91) (was developing in an empty debug directory) and I've got more solid questions/comments:

The current rough working state of this PR is that `mason install` runs `mason install <dep> <version>` for each dependency in a config.json file inside `$MASON_DIR`. That feels incomplete to me:

* These bootstrap scripts usually clone mason themselves, so `.mason/` (`$MASON_DIR`) does not exist before it's run. `mason install` looks for the config.json there — should it be looking in the dir where the command is run? In one dir up from `$MASON_DIR`? (And it should probably actually be called something like `mason_config.json`…)
* A lot of mason bootstrap files have a `dep` command that looks like
```
./.mason/mason install <dep> <version>
./.mason/mason link <dep> <version>
```
Would you always want to `mason install <d> <v> && mason link <d> <v>`? If so, should `mason install` do both? Or should it be another command like `mason configure` (or some other word that implies "install and link from config file")?
* Or should `config.json` handle different batch commands , like
``` json
{
  "install": {
    "cmake": "3.2.2",
    "lua": "5.3.0",
    "luabind": "dev",
    "boost": "1.57.0"
  },
  "link": {
    "cmake": "3.2.2",
    "lua": "5.3.0",
    "luabind": "dev",
    "boost": "1.57.0"
  }
}
```
* Is this just totally a fool's errand? (I say this after I looked at how deep npm's `install` logic gets…)

springmeyer:
I think:

 - `mason install` should look in current directory for the config, so `./mason.json`
 - `mason install` and `mason link` should stay separate for now. (I like the idea of automatically linking but that's probably to much of a departure from how mason works right now). So both `mason install` and `mason link` could respond similarly to the `./mason.json` being found in the current working directory.

> Is this just totally a fool's errand? (I say this after I looked at how deep npm's install logic gets…)

I don't think so, this feels like great progress forward. But let me know what you saw inside npm :). The main thing we'd never want from npm is the possibility of duplicately installed modules / different deps in the tree requiring different versions. But other than that a lot of the install logic perhaps could be more similar...

springmeyer:
Closing. Being able to install from a config file is still often desirable, but not actionable right now.

PR #106, opened 2015-07-28
Adopt Bitcode for iOS

1ec5:
This PR adds `-fembed-bitcode` to iOS 9+ builds.

/ref mapbox/mapbox-gl-native#1957
/cc @kkaefer @incanus


1ec5:
Android builds are failing, but otherwise I think this is good to go.

PR #107, opened 2015-07-29
Fixes for cross compiling on Linux

tmpsantos:
Fix toolchain permissions and recursion when installing the toolchain.


tmpsantos:
@kkaefer please review

PR #109, opened 2015-07-31
Use correct host flag

igorti:
This small `host` flag typo fails the build when compiling for iOS.


springmeyer:
Nice catch, thanks @igorti!

PR #110, opened 2015-08-04
Hard-code minimum iOS SDK version

1ec5:
We should hard-code the minimum iOS SDK version to 7.0 instead of the selected Xcode toolchain’s latest supported SDK version. We already hard-code the minimum OS X SDK version (d43175d51fa4df4a2995ceac7cff7a6560ad01e8).

ref mapbox/mapbox-gl-native#1707

/cc @kkaefer @incanus


kkaefer:
Nice catch.

PR #117, opened 2015-10-05
OS X 10.11 fix: /usr/bin/clang

springmeyer:
I've found that upgrading to El Cap (OS X 10.11) broke my ability to compile C programs like:

```
In file included from cryptlib.c:117:
./cryptlib.h:62:11: fatal error: 'stdlib.h' file not found
# include <stdlib.h>
          ^
1 error generated.
make[1]: *** [cryptlib.o] Error 1
make: *** [build_crypto] Error 1
```

^^ that was seen doing `brew install wget` which needed openssl.

But upgrading to the latest XCode via the app store (v7.0.1) fixed that issue.

However the same error persists now for another reason even with updated Xcode and impacts the viability of Mason on 10.11:

```
configure:20620: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -E -I/Users/dane/projects/mason/postgis-2.1.7/mason_packages/osx-10.11/gdal/1.11.2/include conftest.c
In file included from conftest.c:65:
In file included from /Users/dane/projects/mason/postgis-2.1.7/mason_packages/osx-10.11/gdal/1.11.2/include/gdal.h:42:
/Users/dane/projects/mason/postgis-2.1.7/mason_packages/osx-10.11/gdal/1.11.2/include/cpl_port.h:144:10: fatal error: 'stdio.h' file not found
#include <stdio.h>
         ^
1 error generated.
```

Seen trying to build `postgis` from source with mason.

The above pull request works around this issue for reasons that I can only guess at. Perhaps Apple has broken the `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang` and don't have plans to fix it since `/usr/bin/clang` is preferable. Mason was using `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang` just because I saw xcodebuild using back it the day and figured it was best practice.

### DIFF
--- a/mason.sh
+++ b/mason.sh
@@ -58,8 +58,8 @@ if [ ${MASON_PLATFORM} = 'osx' ]; then
     # to correctly target c++11 for build systems that don't know about it yet (like libgeos 3.4.2)
     # But because LDFLAGS is also for C libs we can only put these flags into LDFLAGS per package
     export LDFLAGS="-Wl,-search_paths_first ${SYSROOT_FLAGS}"
-    export CXX="${MASON_XCODE_ROOT}/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++"
-    export CC="${MASON_XCODE_ROOT}/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang"
+    export CXX="/usr/bin/clang++"
+    export CC="/usr/bin/clang"
 
 elif [ ${MASON_PLATFORM} = 'ios' ]; then
     export MASON_HOST_ARG="--host=arm-apple-darwin"


incanus:
If you are hardcoding the long `clang` path, you shouldn't. Check out `xcode-xselect -p` and/or `xcrun` for  proper path handling.

1ec5:
:+1:

Alternatively, you can run `xcrun clang`, which resolves to the clang inside your `xcode-select`ed Xcode.app without having to worry about toolchain paths. We shouldn't rely on /Applications/Xcode.app being the right Xcode or even existing. At any given time, I have three different Xcodes installed, due to Travis' upgrade cycle, and use `xcode-select` to choose the right one to compile with on the command line.

springmeyer:
@1ec5 and I sat together today and debugged this in more detail. We both could replicate the issue using this simple testcase:


```c
#include <stdio.h>

int main() {

    return 0;
}
```

We saw that `/usr/bin/clang` works:

```
/usr/bin/clang test.c && ./a.out # no errors, no output as expected
```

While `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang` does not:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang test.c && ./a.out
test.c:1:10: fatal error: 'stdio.h' file not found
#include <stdio.h>
         ^
1 error generated.
```

Then we added `-v` to the compiler flags and compared the results. We found the difference is that `/usr/bin/clang` puts:

```
-isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk
```

In the compile flags by default whereas the `clang` inside the `MacOSX.platform` makes no such assumption. Odd, but okay and perhaps an intended in XCode 7.

I confirmed that Mason is already correctly setting `-isysroot` in the `$CFLAGS` so this should allow all build systems to work:

```
$ ~/.mason/mason env CFLAGS
-isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk -arch x86_64 -mmacosx-version-min=10.8
```

However the problem is that some build scripts assume they can use `$CXX` without any custom flags. For example even though `$CFLAGS` contains `isysroot` the libpng build system still breaks at configure time because its configure checks don't use `$CFLAGS`:

```
configure: creating ./config.status
config.status: creating Makefile
config.status: creating libpng.pc
config.status: creating libpng-config
config.status: creating config.h
config.status: executing depfiles commands
config.status: executing libtool commands
rm -f pnglibconf.c pnglibconf.tf[45]
awk -f ./scripts/options.awk out=pnglibconf.tf4 version=search\
	    ./pngconf.h ./scripts/pnglibconf.dfa\
	    ./pngusr.dfa  1>&2
awk -f ./scripts/options.awk out=pnglibconf.tf5 pnglibconf.tf4 1>&2
rm pnglibconf.tf4
mv pnglibconf.tf5 pnglibconf.c
rm -f pnglibconf.out pnglibconf.tf[12]
test -d scripts || mkdir scripts || test -d scripts
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -E -DHAVE_CONFIG_H -I. \
	     -DPNGLIB_LIBNAME='PNG16_0' -DPNGLIB_VERSION='1.6.17' -DSYMBOL_PREFIX='' -DPNG_NO_USE_READ_MACROS -DPNG_BUILDING_SYMBOL_TABLE  pnglibconf.c > pnglibconf.tf1
pnglibconf.c:33:11: fatal error: 'zlib.h' file not found
# include <zlib.h>
          ^
1 error generated.
make: *** [pnglibconf.out] Error 1
```

So, another workaround would be changing Mason to put `$CFLAGS` into the `$CXX` env variable:

```diff
diff --git a/mason.sh b/mason.sh
index 1251421..ec55ee8 100644
--- a/mason.sh
+++ b/mason.sh
@@ -59,7 +59,7 @@ if [ ${MASON_PLATFORM} = 'osx' ]; then
     # But because LDFLAGS is also for C libs we can only put these flags into LDFLAGS per package
     export LDFLAGS="-Wl,-search_paths_first ${SYSROOT_FLAGS}"
     export CXX="${MASON_XCODE_ROOT}/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++"
-    export CC="${MASON_XCODE_ROOT}/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang"
+    export CC="${MASON_XCODE_ROOT}/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang ${CFLAGS}"
 
 elif [ ${MASON_PLATFORM} = 'ios' ]; then
     export MASON_HOST_ARG="--host=arm-apple-darwin"
```

^^ This fixes the libpng build but is not ideal since we know that some old build systems assume that `$CC` and `$CXX` can be safely evaluated as paths/files that exist.

So, after discussing this in detail @1ec5 and I both feel like the workaround of just using `/usr/bin/clang` when targeting OS X (not iOS) is the easiest workaround.

springmeyer:
Restarted travis jobs which failed for an unrelated reason. Will merge this once we are all green.

springmeyer:
pwnvest on. Turns out, despite being sure I had the command line tools installed, I did not. Running `xcode-select --install` brought up a prompt. After installing them I no longer have this problem. I'm a bit in disbelief since this would mean that @1ec5 also was able to replicate because he also did not have the "command line tools" installed for Xcode 7 installed SDK...

1ec5:
Yes, that much makes sense: we both had a stub `/usr/bin/clang` that summons `xcrun clang` – that comes with the system. CLI Tools replaces it with a full-fledged compiler. It’s absolutely correct that this PR switched from requiring Xcode (to be installed at the default location) to requiring either CLI Tools or Xcode (to be installed anywhere). What didn’t occur to me at first is that the OS X SDK wouldn’t be installed anywhere accessible from `/usr/bin/clang` unless CLI Tools were installed, and that you need that SDK in order to compile even the simplest programs with the `clang` inside Xcode.

In short, I’m glad we merged this PR.

PR #125, opened 2015-12-23
Adding support for suffixed version of qmake

danilocesar:
Systems like Fedora23 doesn't ship pure versions of qmake, only
suffixed ones like qmake-qt4 and qmakq-qt5. With this patch we can
make it work for all versions.


kkaefer:
Thanks!

PR #129, opened 2016-03-08
Fix linking for minjur package

springmeyer:
Same fix as #128 (diff package) /cc @rclark


rclark:
:+1: thanks @springmeyer -- not a big deal for how I'm using it, but good to see where the binaries ought to land.

PR #131, opened 2016-03-10
Don't subversion for different releases of OS X

### LINKED ISSUE #118: Support OS X 10.11 (aka El Cap)
Currently travis only builds binaries for OS X 10.10. So anyone upgrading to El Cap will immediately be hit with having to build from source.

Two potential solutions:

1) Start building for both 10.10 and 10.11 on travis

2) Stop versioning on `${MASON_PLATFORM_VERSION}`. In other words stop having binaries use `10.10` in the path such that binaries built on `10.10` will still be found (and will work) on 10.11 (refs https://github.com/mapbox/mason/issues/38#issuecomment-145685563_

2) is vastly better and simple, we just need to think through a way to move on it in a back compatible way.

/cc @mikemorris @1ec5 @incanus @kkaefer

jfirebaugh:
Instead of building for `osx-10.10`, `osx-10.11`, etc., build a single `osx-all` version. Continue to use `-mmacosx-version-min=10.8`.

Fixes #118

Before testing this, I'll need to copy existing `osx-10.10` / `osx-10.11` packages to `osx-all` on S3. @springmeyer @kkaefer, should I copy one or the other, or both? And if both, should the `osx-10.10` or `osx-10.11` package get priority when they both exist?

### DIFF
--- a/mason.sh
+++ b/mason.sh
@@ -40,16 +40,18 @@ esac
 
 if [ ${MASON_PLATFORM} = 'osx' ]; then
     export MASON_HOST_ARG="--host=x86_64-apple-darwin"
-    export MASON_PLATFORM_VERSION=`xcrun --sdk macosx --show-sdk-version`
+    export MASON_PLATFORM_VERSION=`uname -m`
 
-    if [[  ${MASON_PLATFORM_VERSION%%.*} -ge 10 && ${MASON_PLATFORM_VERSION##*.} -ge 11 ]]; then
+    MASON_SDK_VERSION=`xcrun --sdk macosx --show-sdk-version`
+    MASON_SDK_ROOT=${MASON_XCODE_ROOT}/Platforms/MacOSX.platform/Developer
+    MASON_SDK_PATH="${MASON_SDK_ROOT}/SDKs/MacOSX${MASON_SDK_VERSION}.sdk"
+
+    if [[  ${MASON_SDK_VERSION%%.*} -ge 10 && ${MASON_SDK_VERSION##*.} -ge 11 ]]; then
         export MASON_DYNLIB_SUFFIX="tbd"
     else
         export MASON_DYNLIB_SUFFIX="dylib"
     fi
 
-    MASON_SDK_ROOT=${MASON_XCODE_ROOT}/Platforms/MacOSX.platform/Developer
-    MASON_SDK_PATH="${MASON_SDK_ROOT}/SDKs/MacOSX${MASON_PLATFORM_VERSION}.sdk"
     MIN_SDK_VERSION_FLAG="-mmacosx-version-min=10.8"
     SYSROOT_FLAGS="-isysroot ${MASON_SDK_PATH} -arch x86_64 ${MIN_SDK_VERSION_FLAG}"
     export CFLAGS="${SYSROOT_FLAGS}"
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+mason_packages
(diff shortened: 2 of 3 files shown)


jfirebaugh:
Seems like copying just `osx-10.11` is probably the way to go. Proposed command line:

```
aws s3 cp --recursive s3://mason-binaries/osx-10.11 s3://mason-binaries/osx-all
```

springmeyer:
One followup we'll want after this becomes standard is to remove the multiple `matrix` entries for all travis packages building for both 10.10 and 10.11 (to avoid them clobbering each other on upload).

jfirebaugh:
Test build of mapbox-gl-native passed on all platforms:

* [Linux](https://travis-ci.org/mapbox/mapbox-gl-native/builds/115204507)
* [OS X](https://www.bitrise.io/build/edfa82019ef76ea5)
* [iOS](https://www.bitrise.io/build/8f840a35a2e3a044)
* [Android](https://www.bitrise.io/build/c52bb90f3295e278)

mason.sh
```diff
@@ -40,16 +40,18 @@ esac
 
 if [ ${MASON_PLATFORM} = 'osx' ]; then
     export MASON_HOST_ARG="--host=x86_64-apple-darwin"
-    export MASON_PLATFORM_VERSION=`xcrun --sdk macosx --show-sdk-version`
+    export MASON_PLATFORM_VERSION="all"
```
kkaefer: `x86_64` instead? We're only using `all` on iOS because we're producing fat binaries for that platform; all other platforms use arch-specific platform versions.

mason.sh
```diff
@@ -40,16 +40,18 @@ esac
...
 
-    if [[  ${MASON_PLATFORM_VERSION%%.*} -ge 10 && ${MASON_PLATFORM_VERSION##*.} -ge 11 ]]; then
+    MASON_SDK_VERSION=`xcrun --sdk macosx --show-sdk-version`
+    MASON_SDK_ROOT=${MASON_XCODE_ROOT}/Platforms/MacOSX.platform/Developer
+    MASON_SDK_PATH="${MASON_SDK_ROOT}/SDKs/MacOSX${MASON_SDK_VERSION}.sdk"
+
+    if [[  ${MASON_SDK_VERSION%%.*} -ge 10 && ${MASON_SDK_VERSION##*.} -ge 11 ]]; then
```
kkaefer: This will fail for Mac OS 11.0

mason.sh
jfirebaugh: :+1:, I'll switch to `x86_64`.

mason.sh
jfirebaugh: The logic hasn't changed from what was here before, just the variable was renamed.

mason.sh
jfirebaugh: Using `x86_64` causes a failure in one of the tests. I'll need someone to help me figure out why https://travis-ci.org/mapbox/mason/jobs/115412535.

mason.sh
kkaefer: Yup, but let's fix it along the way.

jfirebaugh: Could we just set `MASON_DYNLIB_SUFFIX="tbd"` unconditionally?

kkaefer: That'd mean we can't use Mason on systems older than El Capitan anymore. The two users of `MASON_DYNLIB_SUFFIX` are `libcurl-system` and `zlib-system`. We could update those to do the conditional check?

jfirebaugh: I'm punting on this -- feel free to ticket a followup.

Issue #67, opened 2015-02-03
immutable dependencies


jfirebaugh:
We had several days of build instability in mapbox-gl-native and mapbox-gl-js due to unexpected changes in upstream mason dependencies. In some cases the problem was that a mason branch was deleted while projects still used it, in other cases it was that the build script changed in ways that broke certain platforms. What's common to both cases is that they revealed an architectural weakness in how we manage dependencies with mason: there's no way to test or stage changes to packages without potentially breaking downstream projects.

This is closely related to an architecture invariant that I think every package system comes to adopt (or regret not adopting) at some point: once published, a particular version of a package should _never_ change, period. Being able to overwrite binaries or install scripts for a particular version is just asking for trouble -- namely, uncontrollable breakage of downstream projects.

Let's consider:

* Uploading build / install scripts to S3 along with binaries, so there's no direct dependency on particular git repositories or branches being available.
* Suffixing the package version number with a monotonically increasing "build" number, e.g. mesa-10.4.3-1, mesa-10.4.3-2.
* Requiring that dependencies be specified by package version and build number.
* Never overwriting a package-build version on S3.

This would ensure that we can safely iterate on packages in mason without risking uncontrollable downstream breakage, and downstream packages get reproducible builds and the assurance that dependencies won't change out from underneath them.

ljbade:
@jfirebaugh what about package version + git hash in configure?
Then we can keep an archive of build script + binaries on s3 for each hash?

Suffixing package version will result in an explosion of mason branches.

jfirebaugh:
> Suffixing package version will result in an explosion of mason branches.

I was envisioning keeping the same branch naming conventions and incrementing a build number stored somewhere else. Using a hash (or short hash like `mesa-10.4.3-a36df3`) has the advantage that Travis could autogenerate it, but the disadvantage that you can't quickly tell which version is more recent.

ljbade:
@jfirebaugh Perhaps tags are better method for capturing package releases, perhaps those can have a version appeneded

mikemorris:
I'd be :+1: on an implementation using tags, as it keeps versions a bit more visible than stashing old packages on S3. Whichever way we implement this though, it would be great to fully make this a mason feature. Something like:
- `mason install mesa 10.4.3` install latest tag and pin to that tag somehow
- `mason install mesa 10.4.3 --tag=v1.0.0` install a specific tag (verbose to avoid confusion with package version)
- `mason tags mesa 10.4.3` list tags

[...]

Issue #60, opened 2015-01-28
Culling packages


springmeyer:
Packages that can be deleted:
- [x] nunicode != latest version (1.5.1)
- [x] libpng != latest version

ljbade:
Oh yeah remove the old packages now that I merged updates into mapbox-gl-native.

Are mason packages referenced from anywhere else?

ljbade:
More for culling:
- [ ] glfw
- [ ] jpeg
- [ ] libcurl - need to test 7.40 with boringssl now that I the patch should be in there - or is it coming in 7.41?
- [ ] libuv (once the iOS lipo bug is fixed)
- [ ] mesa
- [ ] openssl
- [ ] sqlite

springmeyer:
:+1: 

> Are mason packages referenced from anywhere else?

only mbgl and mapnik - let's me ruthless in this early stage in cleaning up packages. If things break downstream they likely should be breaking (so we fix).

yhahn:
... 0 tags? ; ) https://github.com/mapbox/mason/tags

ljbade:
@springmeyer which GLFW is the latest one?

springmeyer:
@ljbade @kkaefer would know

springmeyer:
@yhahn packages themselves are the "releases" and so I'd not been planning on mason tags (like homebrew https://github.com/Homebrew/homebrew/releases)

ljbade:
I prefer the branch system as it allows updates to an existing version, such as extending the Android ABIs I did recently.

kkaefer:
@yhahn I thought about using tags, but that means we'd have two version numbers per package, the packaged software's version and the package version, which is probably going to be confusing; the idea is that a package of a software version doesn't need to be versioned; only bufixed, i.e. there should not be any need for a consumer to pin to an "old" package release.

kkaefer:
We can switch to GLFW 3.1, which was released in a few weeks ago and has all the bug fixes we need.

mikemorris:
With https://github.com/mapbox/mapbox-gl-native/pull/796, `mesa` packages prior to `10.4.3` should be clear to be culled.

mikemorris:
From @jfirebaugh:
> did someone delete mason branches that are in use? specifically mesa-10.3.1?
old branches / pull requests are now failing
"old" as in "created 3 days ago"
let's be a little more conservative in deleting mason branches

jfirebaugh:
[...]

PR #133, opened 2016-03-11
Uni-branch

jfirebaugh:
[Test build here](https://travis-ci.org/mapbox/mason/builds/115220596), triggered locally with `./mason trigger jni.hpp 2.0.0`.

It looks as though the triggered build API is rate limited; the responses come back with a `"remaining_requests"` key that starts with a value of 10 and counts down from there. I got down to 1 before getting everything in order. Haven't seen it reset yet, and can't find documentation about what the timeframe is.


jfirebaugh:
Next step here would be to write a script:

```
for $branch in $branches:
   git checkout $branch
   git mv script.sh scripts/${MASON_NAME}/${MASON_VERSION}.sh
   git rm .travis.yml
   git commit -a -m "Preparing for single branch merge"
git checkout triggered-builds
git merge $branches
```

That way we preserve the git history of all the scripts.

springmeyer:
Because some packages have patches and test files I think we should use directories for packages, so perhaps:

```sh
mkdir -p ${MASON_NAME}-${MASON_VERSION}
git mv *.* ${MASON_NAME}-${MASON_VERSION}/
```

jfirebaugh:
Good progress here. Migration script is written and tested -- as you can see, all the commits from all the individual branches are represented here. There's a bug in the script that caused all boost libraries to get folded into a single directory named `boost_lib${BOOST_LIBRARY}`; I'll fix that and redo the merge.

@springmeyer @kkaefer Want to review the last few interesting commits here, and the overall directory structure?

jfirebaugh:
I think this is ready to go. I've fixed the boost directories and brought over the `.travis.yml` from each former branch, and the contents are now used for the triggered build.

The remaining question is the discussion in https://github.com/mapbox/mason/commit/86dbb40a6ea5753e4ea9350b7b37bd1e965a5f4a, which let's move here.

@jfirebaugh
> Now that we're keeping the scripts in a single branch, we could remove this block and the separate .scripts directory entirely, and use the script file from the local checkout instead. This would improve stability in use cases like mapbox-gl-native. The current behavior can produce broken builds of past revisions, for example during git bisects, when the tip version of some script.sh is incompatible with the version of mason / mason.sh that mapbox-gl-native has pinned. Using mason, mason.sh, and a script.sh all from the same checkout would make that much less likely.

@kkaefer
> Hm, that would mean we'd have to re-pin the .mason directory whenever we update a software version?

> Overall, our goal for mason is to split the building packages part from the installing/depending on packages part. E.g. I have a sample implementation of a pure CMake-based package installer that integrates into find_library. That also means that we won't have to download a script.sh file anymore because we can directly download the tarball and error if it doesn't exist.

@jfirebaugh
> > Hm, that would mean we'd have to re-pin the .mason directory whenever we update a software version?

> Yes, and that's a good thing. I don't want the build behavior of past revisions of mapbox-gl-native to change when someone modifies a script.sh. For more on this, see #67. The solution I propose there involves adding build revisions to the version number and writing the script.sh to S3. Using script.sh from the local checkout would be a lighter-weight approach that solves most of the same problems. (The part it doesn't solve is prebuilt binaries getting overwritten on S3.)

jfirebaugh:
Another advantage of using `script.sh` from the pinned local checkout is it will allow us to be more aggressive about culling obsolete package versions (#60) from mason `master` without fear of breaking downstream consumers.

springmeyer:
> Using script.sh from the local checkout would be a lighter-weight approach that solves most of the same problems

:+1: 

> Another advantage of using script.sh from the pinned local checkout is it will allow us to be more aggressive about culling obsolete package versions (#60) from mason master without fear of breaking downstream consumers.

:+1:

PR #135, opened 2016-03-16
Use bfd instead of gold linker on aarch64

tmpsantos:
We were getting segmentation faults and illegal instructions. My take is support for aarch64 + C++14 symbols is fresh. I didn't find anyone else reporting similar issues so I created one on AOP.

AOP issue:
https://code.google.com/p/android/issues/detail?id=204151

Mapbox GL Native issues:
https://github.com/mapbox/mapbox-gl-native/issues/1956
https://github.com/mapbox/mapbox-gl-native/issues/3128


springmeyer:
Curious what version of the gold linker/binutils is available on android? FWIW I hit bugs in gold around `-flto` (https://github.com/Project-OSRM/node-osrm/pull/153#issuecomment-194053521) so I packaged 2.26 with mason (https://github.com/mapbox/mason/blob/master/scripts/binutils/2.26/script.sh) to workaround those bugs.

Issue #141, opened 2016-03-30
Drop support for ~/.mason


jfirebaugh:
We should eliminate `~/.mason` as a default location and document the use of a git submodule as the recommended approach.

Related to #140.

springmeyer:
:+1: mason should be able to be installed anywhere, and we should probably stick with unix conventions in the future: `bin/mason` for the command and `share/mason/` perhaps for scripts? Then people could just put it on PATH or install locally (like with a submodule as you describe). All of mason's logic would be smart and relative.

jfirebaugh:
> eliminate `~/.mason` as a default location

Done in #151.

> document the use of a git submodule as the recommended approach

Still needs to be done.

Issue #140, opened 2016-03-30
Eliminate MASON_DIR


jfirebaugh:
Instead of requiring a `MASON_DIR` environment variable, mason should locate everything using paths relative to the location of the `mason` script.

springmeyer:
Now that scripts are in the same directory tree as mason itself, this should be possible :+1: Will require quite a bit of script cleanup, but we could team up on that.

jfirebaugh:
`MASON_DIR` is handled internally as of #141.

PR #151, opened 2016-04-08
MASON_DIR != ~/.mason

jfirebaugh:
Work toward #140 / #141.

This is mostly rote search and replace. See individual commits for details.

Remaining `~/.mason` stragglers:

* [x] README.md
* [x] [selfupdate](https://github.com/mapbox/mason/blob/30ce49507e0bfc6a7d52c2fc898692416436ac98/mason#L23-L23) -- should we remove this command?
* [x] postgis 2.1.5 and 2.1.7 test.sh. I see this has been fixed in 2.2.2. Prune the older versions?
* [x] postgres 9.4.0, 9.4.1, 9.5.2 test.sh.
* [x] zlib system test.sh.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ Declare these variables first. `MASON_NAME` and `MASON_VERSION` are mandatory. I
 Then, we're loading the build system with
 
 ```bash
-. ~/.mason/mason.sh
+. ${MASON_DIR}/mason.sh
 ```
 
 Next, we're defining a function that obtains the source code and unzips it:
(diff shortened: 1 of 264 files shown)


springmeyer:
> selfupdate -- should we remove this command?

:+1:

jfirebaugh:
The test failure here is due to a subtle order-of-operations change in 30ce49507e0bfc6a7d52c2fc898692416436ac98 that breaks header-only libraries.

Before, `MASON_HEADER_ONLY=true` would run before sourcing `mason.sh` for a second time. The second time `mason.sh` was sourced, [`MASON_PLATFORM_ID`](https://github.com/mapbox/mason/blob/fb191bb185cbd027a46bdc104123e7535aa9e834/mason.sh#L244) was set to the correct value.

After these changes, `mason.sh` is only sourced the first time, and `MASON_HEADER_ONLY=true` doesn't get set in time to affect `MASON_PLATFORM_ID`.

jfirebaugh:
I switched to a less invasive approach that doesn't change the order of operations. I think this is ready to 🚢  -- testing downstream now.

PR #158, opened 2016-04-21
Add shelf-pack-cpp 1.0.0 to mason

bhousel:
I think I got this right..
👀 @jfirebaugh ?


jfirebaugh:
LGTM. You can run `./mason trigger shelf-pack 1.0.0` with this branch checked out to check that it works.

PR #159, opened 2016-04-22
libosmium@2.6.1

rclark:
Remaining question is whether or not to include the protozero headers that come bundled with libosmium

cc @springmeyer


springmeyer:
Per chat protozero should be packaged separately - will be great to have this, thanks @rclark!

rclark:
Now this includes libosmium, protozero, and utfcpp as standalone mason packages. Look good @springmeyer?

springmeyer:
Looking great @rclark - thank you.

PR #162, opened 2016-04-29
Introduce [suffix] optional parameter

### LINKED ISSUE #157: Plan for dealing with libstdc++ ABI break
The [libstdc++ ABI break](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html) poses a challenge for mason (https://github.com/mapbox/mason/issues/155) as we see more use of gcc/libstdc++ >= 5.1.

This affects any library that exports a symbol that uses an ABI-tagged type -- e.g. a function returning `std::string`.

As I understand it, if mason wants to package libraries of this type, then the choices are:

* Require downstream consumers to compile everything with libstdc++ < 5.1, or with `-D_GLIBCXX_USE_CXX11_ABI=0`. This is the status quo.
* Require downstream consumers to compile everything with libstdc++ >= 5.1 and `-D_GLIBCXX_USE_CXX11_ABI=1`. This is the "force everyone to upgrade" option.
* Provide dual-ABI symbols in each library itself. This is the "provide backward compatibility by copying what libstdc++ itself does" option. It's probably infeasible due to the amount of work on 3rd party libraries that would be required.
* Start embedding the ABI setting in the mason _platform_ slug (e.g. `linux-x86-cxx03abi` vs `linux-x86-cxx11abi`). Build all libraries for both ABIs.
* Start embedding the ABI setting in the mason _package_ slug (e.g. `gtest-1.7.0-cxx03abi` vs `gtest-1.7.0-cxx11abi`).  Build all libraries that export ABI-tagged symbols for both ABIs.

I currently favor the last option but would like to hear other opinions.

cc @springmeyer @tmpsantos

tmpsantos:
Append a suffix on any package, so you can build/install
different binaries of packages with the same version like:
```
foobar-1.0-cxx11abi
foobar-1.0-gcc4.9
foobar-1.0-patch1
foobar-1.0-whatever
```
Normal usage remains:
```
$ mason build foobar 1.0
$ mason publish foobar 1.0
```
Or:
```
$ mason build foobar 1.0 cxx11abi
$ mason publish foobar 1.0 cxx11abi
```
Fixes #157

### DIFF
--- a/mason
+++ b/mason
@@ -3,12 +3,13 @@
 MASON_COMMAND=$1 ; shift
 MASON_NAME=$1 ; shift
 MASON_VERSION=$1 ; shift
+MASON_SUFFIX=$1 ; shift
 
 set -e
 set -o pipefail
 
 function usage {
-    echo "Usage: $0 <command> <lib> <version>"
+    echo "Usage: $0 <command> <lib> <version> [suffix]"
 }
 
 if [ -z "${MASON_COMMAND}" ]; then
@@ -37,6 +38,10 @@ elif [ -z "${MASON_VERSION}" ]; then
     usage
     echo "Missing <version>"
     exit 1
+fi
+
+if [ ! -z "${MASON_SUFFIX}" ]; then
+    export MASON_SUFFIX="-${MASON_SUFFIX}"
 fi
 
 if [ "${MASON_COMMAND}" = "trigger" ]; then
--- a/mason.sh
+++ b/mason.sh
@@ -246,8 +246,9 @@ if [[ ${MASON_HEADER_ONLY} == true ]]; then
 else
     MASON_PLATFORM_ID=${MASON_PLATFORM}-${MASON_PLATFORM_VERSION}
 fi
-MASON_PREFIX=${MASON_ROOT}/${MASON_PLATFORM_ID}/${MASON_NAME}/${MASON_VERSION}
-MASON_BINARIES=${MASON_PLATFORM_ID}/${MASON_NAME}/${MASON_VERSION}.tar.gz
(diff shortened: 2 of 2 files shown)


tmpsantos:
:eyes: @jfirebaugh

jfirebaugh:
Can we have the suffix be part of the version instead? `1.0-cxx11abi`, `1.0-gcc4.9`, etc. I think this will work already without any changes to mason internals.

tmpsantos:
I tried that first, but that would require to create a new suffixed copy of
the package script instead of reusing the existing one. Would that be OK?

.../gtest/1.7.0/
.../gtest/1.7.0-cxx11abi/

jfirebaugh:
Yeah, I think that's fine. In the long term we should plan on transitioning to building new packages and new versions of existing packages with `cxx11abi` by default, and without a suffix, so the need for separate scripts will decrease over time.

tmpsantos:
@springmeyer @jfirebaugh what should I do about this one? Should we support arbitrary prefixes or copy and paste packages scripts for adding prefixes and keep mason scripts unchanged?

I'm fine with both. Asking again because of:
https://github.com/mapbox/mason/issues/157#issuecomment-215906183

tmpsantos:
To be clear, the alternative suggested would be this https://github.com/mapbox/mason/commit/5a52cb782e344b630409a0b36fb52192079f19a4

jfirebaugh:
Let's go with creating custom package scripts as suggested in https://github.com/mapbox/mason/issues/157#issuecomment-215906183 and not make mason internal changes.

tmpsantos:
> Let's go with creating custom package scripts as suggested in #157 (comment) and not make mason internal changes.

:+1:

PR #175, opened 2016-05-28
add leveldb@1.18

morganherlocker:
This PR adds [leveldb](https://github.com/google/leveldb) as a header-only library. This is my first time publishing a mason lib, so I would love to get a review when you have a minute @springmeyer to make sure I am doing this correctly.

cc @rclark

### DIFF
--- a/scripts/leveldb/1.18/.travis.yml
+++ b/scripts/leveldb/1.18/.travis.yml
@@ -1,0 +1,22 @@
+language: cpp
+
+sudo: false
+
+matrix:
+  include:
+    - os: osx
+      compiler: clang
+    - os: linux
+      compiler: clang
+
+env:
+  global:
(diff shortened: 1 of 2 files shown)


springmeyer:
@morganherlocker I tried building and hit:

```
$ ./mason build leveldb 1.18
* Downloading https://github.com/google/leveldb/archive/v1.18.tar.gz...
Hash d90b5cadb7a366a2ab27ec8b5ed1ea9445c9a2df of file /Users/dane/.mason/mason_packages/.cache/leveldb-1.18 doesn't match 27d2e6530859d9d86f54ed27d8538b781e573af2
```

That fixed the hash, but then the directory unpacked was `google-leveldb-803d692` because that is how tarball urls from github get named. So I switched the url to the release download and then things worked great. I think with those changes this is ready to merge. Then you need to generate a `TRAVIS_TOKEN` and then you can publish the package locally like:

```
./mason trigger leveldb 1.18
```

Issue #69, opened 2015-02-11
Broken: Symlinking to install


springmeyer:
The documented installation instructions (https://github.com/mapbox/mason#installation) actually don't work. You'll get something like:

```sh
~/tmp 》mason install libzip 0.11.2
/usr/local/bin/mason: line 54: /usr/local/bin/mason.sh: No such file or directory
```

I presume this broke in 
https://github.com/mapbox/mason/commit/ad3dd780fbfe209042b5ebe96c2e74dcb7659052#diff-5c29c2e513aadfe372fd0af7553b5a6c and test coverage for this is missing since we don't symlink on travis and also set MASON_DIR: https://github.com/mapbox/mason/blob/master/.travis.yml#L20.

rclark:
How out of the question would it be to remove mason's dependence on living in `~/.mason` entirely?

I'm considering how I might include mason as a dependency in a `package.json` file, and how I might utilize it in something like a `postinstall` script to load up binary dependencies.

jfirebaugh:
Related to #140 and #141.

TobWen:
You should remove the installation howto from https://github.com/mapbox/mason since it also contain the linkage. With the current bash script to detect the mason's root directory, many users will run into the same problem again and again.

springmeyer:
Fixed with #198

PR #177, opened 2016-05-31
Fixing error when mason is a symlink

danilocesar:
At last on Fedora, ROOTDIR var ends up in /usr/local/bin when mason
is a symlink, and then it fails to call other sub-scripts.

In case readlink fails, it uses BASH_SOURCE[0] as fallback.

* It's tested on Linux, I don't own a OSX machine to test it there...

### DIFF
--- a/mason
+++ b/mason
@@ -16,7 +16,9 @@ if [ -z "${MASON_COMMAND}" ]; then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Refs #69. Ultimately we need a solution for this that also works on OS X, where `readlink` does not have an `-f` option.

springmeyer:
Fixed in #198, and alternative submodule approach documented in #192

PR #180, opened 2016-06-01
Remove armv7s as an iOS target architecture

friedbunny:
This pull request stops Mason from building libraries with `armv7s` slices for the iOS platform. This primarily affects geojson-vt.

The Mapbox iOS SDK removed support for the `armv7s` architecture in https://github.com/mapbox/mapbox-gl-native/issues/4704 after we determined it was unnecessary. Devices that rely on `armv7s` can safely fall back to `armv7` with, in our case, no measurable performance impact.

/cc @jfirebaugh @springmeyer @1ec5

PR #181, opened 2016-06-01
Stop building armv7s for iOS

1ec5:
This change removes the `armv7s` slice from iOS build products.

We decided in mapbox/mapbox-gl-native#4704 to drop the armv7s slice from the Mapbox iOS SDK, because iPhone 5 and 5c fall back to `armv7` and the performance boost of `armv7s` over `armv7` is negligible. However, libgeojsonvt.a’s armv7s slice still gets lipo’d into the SDK’s static library.

/cc @friedbunny @jfirebaugh

### DIFF
--- a/mason.sh
+++ b/mason.sh
@@ -414,7 +414,7 @@ function mason_build {
     if [ ${MASON_PLATFORM} = 'ios' ]; then
 
         SIMULATOR_TARGETS="i386 x86_64"
-        DEVICE_TARGETS="armv7 armv7s arm64"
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Duplicate of #180 right?

friedbunny:
Whoops, shoulda' tagged @1ec5 in #180. 😁

1ec5:
lol, you beat me to the punch, and I like your commit message better. 😄

PR #185, opened 2016-06-07
Build + testing improvements

### LINKED ISSUE #184: C++11 test build failing in master
The compiler needs to be properly upgraded: https://travis-ci.org/mapbox/mason/jobs/135978898

Will work on fixing this.

springmeyer:
This moves to:

  - Testing on both Linux and OS X
  - Avoids excessive usage of multiple jobs, collapses to running all tests in one job
  - Properly installs our target C and C++ compiler in a consistent method:
     - Clang 3.8.0
     - Via mason itself
  - ~~Adds a `package_template.yml` that will be used as the default `.travis.yml` for triggered builds if one does not exist for a given script~~

Fixes #184 

/cc @kkaefer @jfirebaugh

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,24 @@
-language: cpp
+language: generic
 
-sudo: false
-
-env:
-  matrix:
-    - MASON_TEST="unit"
-    - MASON_TEST="c_install"
-    - MASON_TEST="c_build"
-    - MASON_TEST="c_install_symlink_includes"
-    - MASON_TEST="cpp11_header_install"
-    - MASON_TEST="cpp11_install"
-    - MASON_TEST="cpp11_build"
-    - MASON_TEST="c_build_android"
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode7.3
+      compiler: clang
+    - os: linux
+      sudo: false
+      addons:
+        apt:
+          sources:
+           - ubuntu-toolchain-r-test
+          packages:
+           - libstdc++6
+           - libstdc++-5-dev
 
 install:
-# NOTE: must be absolute path for predictable behavior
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Latter two commits look good. I'm not totally sold on the `package_template.yml` approach yet though. What percentage of packages would be able to use this? My impression was that a significant number required some customization to .travis.yml, such as adding iOS or Android targets to the build matrix.

springmeyer:
I removed the `package_template.yml` work from this branch for now, and rebased. Will revisit that work later once I've taken a broader look at the problem.

kkaefer:
It looks like forcing Travis configuration breaks some things, e.g. https://travis-ci.org/mapbox/mason/builds/136389871. Nowhere in my Travis YML file does it say that I want Clang 3.8, but it gets forced on me nonetheless.

kkaefer:
It appears that the first two builds are automatically added somehow, and some of them are failing.

kkaefer:
Same with https://travis-ci.org/mapbox/mason/builds/136396716

jfirebaugh:
The Travis Triggered Build API, which is what `mason trigger` uses, does a merge operation between the `.travis.yml` that you post to the API and the default `.travis.yml` in the repository. So for instance now that the default `.travis.yml` does `source utils/toolchain.sh` in the `install` section, that will automatically be run on any build triggered with `mason trigger` where the package-specific `.travis.yml` does not itself have an `install` section.

This merge behavior is confusing if you aren't aware of it, but could actually be kind of useful to get the sort of shared configuration that @springmeyer was aiming for with `package_template.yml`.

springmeyer:
Oh my I had no idea of that behavior @jfirebaugh thanks for explaining.

Issue #188, opened 2016-06-13
Provide mason linux-x86_64 binaries for clang-tidy 3.8


brunoabinader:

springmeyer:
clang-tidy is available as part of the `clang` v3.8.0 package. I've been thinking that if there is a need for just sub-parts of that fairly large package, then we could re-create, for example, the `clang-tidy` package to just pull from the clang v3.8.0 binaries and repackage just the clang-tidy tool. It should be statically linked and so will work standalone.

brunoabinader:
@springmeyer it seems `clang-format` comes from the same repository as `clang-tidy` - it could be useful to have a binary package of it for e.g. https://github.com/mapbox/mapbox-gl-native/issues/5342.

springmeyer:
@brunoabinader - sure, sounds good. Should be easy to do a `clang-format` package similar to https://github.com/mapbox/mason/pull/189/files. Want to ticket that?

brunoabinader:
> Want to ticket that?

👉 #190 

Super awesome! Thanks @springmeyer 😄

brunoabinader:
Fixed in https://github.com/mapbox/mason/pull/189.

PR #189, opened 2016-06-13
Add clang-tidy v3.8.0 as repackaged binary from parent clang build

springmeyer:
This creates a clang-tidy package that re-packages just the `clang-tidy` binary. You can use `clang-tidy` from the main `clang` v3.8.0 package, but when users need just the minimal `clang-tidy` binary this package is much smaller.

It dynamically determines its version and name from the directory. The idea here is to make it easy to copy and modify for re-packaging other clang binaries in the future.

Refs #188


brunoabinader:
💯 @springmeyer !

Issue #30, opened 2014-11-17
Build sqlite-3.8.6 with -fPIC


mikemorris:
Needed for downstream multithreaded usage of [mapbox-gl-native](https://github.com/mapbox/mapbox-gl-native)

/cc @mapbox/gl

Issue #26, opened 2014-11-10
Build libpng with -fPIC


mikemorris:
Needed for downstream multithreaded usage of [mapbox-gl-native](https://github.com/mapbox/mapbox-gl-native)

/cc @mapbox/gl

PR #193, opened 2016-06-22
geojson-vt-cpp@4.1.2-cxx11abi configure --with-pic

mikemorris:
`-fPIC` alone isn't sufficient?

I think this should fix the following linker error I'm seeing when building [mapbox-gl-native](https://github.com/mapbox/mapbox-gl-native/) with `make node`, but `--with-pic` doesn't look like it's actually getting passed through to the right place.

```
/usr/bin/ld: /home/mikemorris/mapbox-gl-native/mason_packages/linux-x86_64/geojsonvt/4.1.2-cxx11abi/lib/libgeojsonvt.a(geojsonvt.o): unrecognized relocation (0x2a) in section `.text'
/usr/bin/ld: final link failed: Bad value
```

Prior art in #26, #30 - it looks like setting `-fPIC` explicitly in the scripts may be redundant with https://github.com/mapbox/mason/commit/2a955a1650fe65bf31e5a198972244541693e897, but `--with-pic` may still be required.

/cc @springmeyer @kkaefer

### DIFF
--- a/scripts/geojsonvt/4.1.2-cxx11abi/script.sh
+++ b/scripts/geojsonvt/4.1.2-cxx11abi/script.sh
@@ -24,7 +24,7 @@ function mason_compile {
     ln -s ${MASON_DIR} .mason
 
     # build
-    INSTALL_PREFIX=${MASON_PREFIX} ./configure
+    INSTALL_PREFIX=${MASON_PREFIX} ./configure --with-pic
     CXXFLAGS="-fPIC ${CFLAGS:-} ${CXXFLAGS:-}" make install
 }
 


mikemorris:
Oh, could it be possible that geojsonvt simply [isn't passing `CXX_FLAGS` through to gyp](https://github.com/mapbox/geojson-vt-cpp/blob/4d5343ee92e01a9b5cf332b5a90d1dc5788fe2e5/geojsonvt.gyp#L40-L58
)? 
Is something like https://github.com/mapbox/mapbox-gl-native/blob/5832f63acc40fd3aa40065a8e4478af211d18fba/mbgl.gypi#L20 necessary?

mikemorris:
Ugh, yep, https://github.com/mapbox/geojson-vt-cpp/pull/46 fixes this.

PR #194, opened 2016-06-23
libleveldb

morganherlocker:
@springmeyer We added leveldb@1.18 a few weeks ago. I finally got around to using it and realized that it actually was not a header-only lib as I first thought. This PR adds a compile step.

A couple questions:

- `MASON_LIB_FILE=lib/libleveldb.a` is the MASON_LIB_FILE referring to the location of the compiled archive after building, or does it refer to some relative path where mason will place the archive? (if it is the former, this line should probably change to `MASON_LIB_FILE=out-static/libleveldb.a`)
- Leveldb does not use pig-config, but I know the ldflags needed are `-lleveldb -lsnappy`. Is echo'ing this manually the correct way to do this in `mason_ldflags`?
- Do we need to copy the compiled lib archive somewhere like we did for the headers?


springmeyer:
looking correct. But yes we need to run 'make install' for mason to be able to package things are predictable locations and then the /lib directory will be right.

springmeyer:
Seeing leveldb lacks an install target (https://github.com/google/leveldb/pull/2). I'm working on fixing this in the script.sh, will push in a sec.

morganherlocker:
For the same reasons outlined in the [snappy PR](https://github.com/mapbox/mason/pull/196#issuecomment-228235679) I assigned an arbitrary `2.0.0` version for this library as well, since the release versions are not reliable.

PR #195, opened 2016-06-23
adding hpp-skel example project

mapsam:
This adds [hpp-skel](https://github.com/mapbox/hpp-skel) to Mason packaging - starting with version `0.0.1`

cc @springmeyer


mapsam:
@springmeyer I've tagged `0.0.1` for hpp-skel and updated the scripts directory accordingly. One question: I noticed the `script.sh` builds the download URL with an appended `v` for the version. When I try to download hpp-skel with the following URL:

```
https://github.com/mapbox/hpp-skel/archive/v0.0.1.tar.gz
```

... it gives me a 404. The published download works without the v, though. 

```
https://github.com/mapbox/hpp-skel/archive/0.0.1.tar.gz
```

Does Mason require the annotation to include `vX.X.X` for each release?

mapsam:
@springmeyer I've updated the git tag to `v0.0.1` so we should be good here!

springmeyer:
@mapsam 🚢

PR #196, opened 2016-06-24
add snappy

morganherlocker:
This PR adds the snappy compression library. This is a dependency for libleveldb.

cc @springmeyer @camilleanne


morganherlocker:
For this library, I was able to follow the default mason pattern for the most part. The exception to this was versioning. The latest tagged snappy release is from last year and a bunch of improvements have been made on the master branch, including cross platform compiler changes that allow snappy to build on OSX. 

Instead of using a release tarball, I configured this script to download a tar.gz of the latest commit on master, and I set the version number to an arbitrary `2.0.0`. Is there a better way to do this?

springmeyer:
@morganherlocker - the convention I've been using is to name it after the git hash, so `snappy/32d6d7d` and you'd install like `mason install snappy 32d6d7d`.

PR #198, opened 2016-06-24
Fix and test symlinked install

springmeyer:
Solves #69 in a cross platform way.

refs #177

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,4 +27,8 @@ install:
   fi
 
 script:
-- ./test/all.sh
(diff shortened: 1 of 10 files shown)


springmeyer:
👀 @jfirebaugh - looks like `readlink -f` is not needed to get this working. Therefore just using `readlink` solves the problem on both linux and osx. Any concerns merging this?

springmeyer:
This mysteriously broke MASON_DIR setting, reverting for now.

PR #199, opened 2016-06-24
Add geojson@0.1.2-cxx03abi

yhahn:
cc @springmeyer

### DIFF
--- a/scripts/geojson/0.1.2-cxx03abi/script.sh
+++ b/scripts/geojson/0.1.2-cxx03abi/script.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+LIB_VERSION=0.1.2
+
+MASON_NAME=geojson
+MASON_VERSION=${LIB_VERSION}-cxx03abi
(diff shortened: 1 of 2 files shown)


scripts/geojson/0.1.2-cxx03abi/script.sh
```diff
@@ -0,0 +1,41 @@
...
+    cp -r include/mapbox ${MASON_PREFIX}/include/mapbox
+    mv build/libgeojson.a ${MASON_PREFIX}/lib
+}
+
+function mason_cflags {
+    echo -I${MASON_PREFIX}/include
+}
+
+function mason_ldflags {
+    :
```
springmeyer: This could provide `echo -L${MASON_PREFIX}/lib -lgeojson` however that is not critical as the expectation is that clients will be asking for the `static_libs` which offers the same thing, but in a better expression for static libraries because passing to the link the '/full/path/libgeojson.a` is more robust than `-L/path/to -lgeojson`  since the latter might match a libgeojson.so if it existed.

springmeyer:
Looks good to merge

PR #209, opened 2016-07-08
Add gtest 1.7.0-cxx03abi & let no suffix mean cxx11

yhahn:
@jfirebaugh this changes the meaning of the current `gtest 1.7.0` package -- is that ok?


jfirebaugh:
I think this would break git bisects of mapbox-gl-native on linux. Let's instead add new `gtest-1.7.0-1` and `gtest-1.7.0-1-cxx03abi` versions.

yhahn:
:ok_hand:

yhahn:
:weary: so it looks a lot like gcc 5 + `D_GLIBCXX_USE_CXX11_ABI=1` is _not_ working. The build output looks good, sample output:

```
/bin/bash ./libtool --tag=CXX   --mode=compile g++-5 -DHAVE_CONFIG_H -I. -I./build-aux  -I. -I./include  -pthread -DGTEST_HAS_PTHREAD=1 -fPIC -std=c++11 -D_GLIBCXX_USE_CXX11_ABI=1 -c -o src/gtest-all.lo src/gtest-all.cc
```

(https://travis-ci.org/mapbox/mason/jobs/143967106#L634)

But ransacking the `gtest.a` file for cxx11 symbols yields no results:

```sh
readelf -Ws lib/libgtest.a | c++filt | grep ::__cxx11
```

Any clues/questions I should be asking next?

cc @tmpsantos @springmeyer

jfirebaugh:
Does gtest actually export symbols that we would expect to be ABI-tagged?

yhahn:
@jfirebaugh I'm assuming because we're seeing a segfault with these binaries. Unfortunately I'm seeing the same behavior with the new geojson build as well (and can confirm that @tmpsantos's local build has ABI-tagged symbols).

tmpsantos:
> gcc 5 + D_GLIBCXX_USE_CXX11_ABI=1 is not working

Try simply building it on Ubuntu 16.04 with no extra flags and you are good. If you don't want to upgrade your box just for that, you could easily do it in a LXC.

jfirebaugh:
In that case I suspect a libstdc++ versioning issue on Travis. Do we need `libstdc++-6-dev` or some such?

springmeyer:
We've not figured out how to get the `D_GLIBCXX_USE_CXX11_ABI=1` flag working on travis such that cxx11abi packages no longer need to be published locally. Let's track next steps (including travis publish of gtest over at https://github.com/mapbox/mason/issues/319)

PR #211, opened 2016-07-20
export MASON_ROOT

kkaefer:
Dependent invocations of `mason` are currently writing into their own `mason_packages` folder, which makes those builds very slow because they can't reuse the packages already in the master `mason_packages` folder.

This change `export`s the `MASON_ROOT` variable so that all dependent `mason` builds use the root folder.

/cc @springmeyer

### DIFF
--- a/mason.sh
+++ b/mason.sh
@@ -2,7 +2,7 @@ set -e
 set -o pipefail
 # set -x
 
-MASON_ROOT=${MASON_ROOT:-`pwd`/mason_packages}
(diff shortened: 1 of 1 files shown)


springmeyer:
:+1 re-using the same `./mason_packages` sounds good. I've already had several build scripts doing this by doing https://github.com/mapbox/mason/blob/07d2aeb6e25901783ba98387e80f713a8aa15afb/scripts/gdal/2.0.2/script.sh#L20 in the `mason_prepare_compile` step.

PR #212, opened 2016-07-20
Add American Fuzzy Lop

kkaefer:
[american fuzzy lop](http://lcamtuf.coredump.cx/afl/) is a program for fuzzing input data.


friedbunny:
Great name. 🐰

PR #214, opened 2016-07-24
Add libcrypto 1.0.1p

tmpsantos:
The cryptographic library from the OpenSSL project.

https://www.openssl.org/docs/manmaster/crypto/crypto.html

### DIFF
--- a/scripts/libcrypto/1.0.1p/.travis.yml
+++ b/scripts/libcrypto/1.0.1p/.travis.yml
@@ -1,0 +1,30 @@
+language: cpp
+sudo: false
+
+os:
+  - linux
+
+addons:
+  apt_packages:
+    - xutils-dev
+env:
+  global:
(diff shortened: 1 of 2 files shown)


tmpsantos:
@springmeyer would like to hear from you here. The OpenSSL package we make currently only exposes as "LIBS" the `libssl.a`. This is maybe technically correct, but for instance if you use `libcurl` that is built with `openssl`, the `libcrypto` symbols are unresolved.

Seems like `mason` has a hard assumption that a package provides a single library (or I'm missing something).

springmeyer:
@tmpsantos I think I'm not completely following your question. But I'll respond for now with what I do understand:

 - By default the result of `mason ldflags <package> <version>` is the result of `pkg-config <package> --static --libs`: https://github.com/mapbox/mason/blob/47b1f2c96a19fa08de15acc06ac2e373fc82e1bf/mason.sh#L533-L537. It looks like the openssl package as well as this libcrypto pull are depend on that default. 
  - Mason has a validation step that requires that `MASON_LIB_FILE` exists. So if a package provides more than one lib you need to set `MASON_LIB_FILE` to just one of them that exists.

PR #176, opened 2016-07-26
Include configuration variables in package

kkaefer:
When generating a package, we should generate the configuration required to link a package and put it into a configuration file that we ship with the package, for example INI-style like this:

```ini
package=libuv
version=1.7.6
platform=osx-x86_64
cc=-isystem {prefix}/include
cxx=-isystem {prefix}/include
link={prefix}/lib/libuv.a -lpthread -ldl
```

The reason for using INI over a more complicated format is that it's way easier to parse with CMake. It also means that we don't have to invoke `pkg-config` or other build-systems to obtain the flags.

/cc @springmeyer


kkaefer:
This PR changes a few things:

- Add a `config` command that can be used like this: `mason config <package> <version>`. It prints an .ini-style file to the console that looks like this:

    ```ini
name=glfw
version=3.1.2
platform=osx
platform_version=x86_64
include_dirs={prefix}/include
ldflags=-framework Cocoa -framework OpenGL
static_libs={prefix}/lib/libglfw3.a
```

  There are also `options` and `definitions`, but they are omitted when empty. Header only packages have `header_only=true` and omit the `platform_*` keys. The string `{prefix}` is replaced with the installation location of the package after reading the ini file. You can use ini-style comments (starting with `;`). Sections are allowed as well, but are currently ignored.

  When building a package, that file is saved as `mason.ini` into the package so that the build system can retrieve the configuration from it without having to invoke Mason.

- Adds a CMake file that downloads, and unpacks packages, and obtains options via the `mason.ini` file. If a legacy package doesn't have a `mason.ini` file, the CMake file generates one (this requires Mason to be present); we can remove that part of the code once all of our packages have mason.ini files. Use in your package could look like this:

  ```cmake
cmake_minimum_required(VERSION 3.2)

include(.mason/mason.cmake)

mason_use(earcut VERSION 0.11 HEADER_ONLY)
mason_use(libpng VERSION 1.6.20)

add_executable(main main.cpp)

target_add_mason_package(main PRIVATE earcut)
target_add_mason_package(main PRIVATE libpng)
```

- Instead of `file(DOWNLOAD ...)`, we are using `curl` as a subprocess because some versions of CMake out there don't include SSL/TLS support.

- The CMake script auto-detects the platform. When you want a different one (e.g. for iOS), you should set `MASON_PLATFORM=ios`.

mason.cmake
```diff
@@ -0,0 +1,216 @@
...
+
+# Path to Mason executable
+if (NOT MASON_COMMAND)
+    set(MASON_COMMAND "${CMAKE_SOURCE_DIR}/.mason/mason")
+endif()
+
+# Determine platform
+if(NOT MASON_PLATFORM)
+    if (CMAKE_HOST_SYSTEM_NAME STREQUAL "Darwin")
+        set(MASON_PLATFORM "macos")
```
springmeyer: It's currently `osx` not `macos` right?

mason.cmake
```diff
@@ -0,0 +1,216 @@
...
+endif()
+
+# URL prefix of where packages are located.
+if (NOT MASON_REPOSITORY)
+    set(MASON_REPOSITORY "https://mason-binaries.s3.amazonaws.com")
+endif()
+
+# Path to Mason executable
+if (NOT MASON_COMMAND)
+    set(MASON_COMMAND "${CMAKE_SOURCE_DIR}/.mason/mason")
```
springmeyer: Any way this could be relative to avoid needed to specifically check out into `.mason`? Like `set(MASON_COMMAND "./mason")`

mason.sh
```diff
@@ -119,7 +119,16 @@ elif [ ${MASON_PLATFORM} = 'linux' ]; then
...
+    case "${MASON_PLATFORM_VERSION}" in
+        arm-v5-9) export MASON_ANDROID_ABI=arm-v5 ;;
+        arm-v7-9) export MASON_ANDROID_ABI=arm-v7 ;;
+        arm-v8-21) export MASON_ANDROID_ABI=arm-v8 ;;
+        x86-9) export MASON_ANDROID_ABI=x86 ;;
+        x86_64-21) export MASON_ANDROID_ABI=x86 ;;
+        mips-9) export MASON_ANDROID_ABI=mips ;;
+        mips64-21) export MASON_ANDROID_ABI=mips64 ;;
+        *) export MASON_ANDROID_ABI=${MASON_ANDROID_ABI:-arm-v7}
+    esac
```
springmeyer: Could also be accomplished with `if [[ $${MASON_PLATFORM_VERSION} =~ "mips-" ]]; then export MASON_ANDROID_ABI=mips .. `

springmeyer:
@kkaefer Saw that error handling could be improved for the case where a package is not installed yet:

```
~/.mason[176-ini-files]$ ./mason config libpng 1.6.20
/Users/dane/.mason/scripts/libpng/1.6.20/script.sh: line 47: /Users/dane/.mason/mason_packages/osx-x86_64/libpng/1.6.20/bin/libpng-config: No such file or directory
/Users/dane/.mason/scripts/libpng/1.6.20/script.sh: line 48: /Users/dane/.mason/mason_packages/osx-x86_64/libpng/1.6.20/bin/libpng-config: No such file or directory
/Users/dane/.mason/scripts/libpng/1.6.20/script.sh: line 40: /Users/dane/.mason/mason_packages/osx-x86_64/libpng/1.6.20/bin/libpng-config: No such file or directory
/Users/dane/.mason/scripts/libpng/1.6.20/script.sh: line 41: /Users/dane/.mason/mason_packages/osx-x86_64/libpng/1.6.20/bin/libpng-config: No such file or directory
No static library file '/Users/dane/.mason/mason_packages/osx-x86_64/libpng/1.6.20/lib/libpng.a'
name=libpng
version=1.6.20
platform=osx
platform_version=x86_64
~/.mason[176-ini-files]$ echo $?
0
```

I think `./mason config` should exit as soon as one of the sub commands fails.

mason.cmake
kkaefer: Yes. We are setting `macos` because we expect that upstream providers/users also use `macos`, but we change it to `osx` a little further down.

mason.cmake
kkaefer: The intention of this variable is that you can override it in the parent CMake script or as a command line option when running cmake, e.g.: `cmake .. -DMASON_COMMAND=/path/to/mason`

kkaefer:
@springmeyer Looks like `local FOO=$(...)` eats exit codes, so the `set -e` didn't exit the script.

PR #144, opened 2016-08-12
Don't subversion for different releases of iOS

kkaefer:
Is there a reason to require building separate binaries for ios-9.0, ios-9.1, ios-9.2, etc? Like in #118 / #131 for OS X, this seems to provide little benefit but means that our build matrix must explicitly build for multiple iOS versions (see e.g. https://github.com/mapbox/mapbox-gl-native/issues/4546 and https://github.com/mapbox/mason/commit/cbe97886ec604018ea8b0faf1cf28658125e162d).

cc @springmeyer @kkaefer @1ec5

### DIFF
--- a/mason.cmake
+++ b/mason.cmake
@@ -26,10 +26,7 @@ endif()
 
 # Determine platform version string
 if(MASON_PLATFORM STREQUAL "ios")
-    execute_process(
-        COMMAND xcrun --sdk iphoneos --show-sdk-version
-        OUTPUT_VARIABLE MASON_PLATFORM_VERSION
-        OUTPUT_STRIP_TRAILING_WHITESPACE)
+    set(MASON_PLATFORM_VERSION "8.0") # Deployment target version
 elseif(MASON_PLATFORM STREQUAL "android")
     if (ANDROID_ABI STREQUAL "armeabi")
         set(MASON_PLATFORM_VERSION "arm-v5-9")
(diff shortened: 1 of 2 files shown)


1ec5:
As long as the package is built against a newer SDK (preferably the current one), it doesn't matter if it deploys back to iOS 7, at least in terms of Bitcode.

kkaefer:
@1ec5 yeah; with this patch, all packages are built with a min version of 8.0, against the current SDK (9.3). At some point, we can choose to update the min version, e.g. once the SDK doesn't support it anymore, or when we are using features that aren't in 8.0 yet.

1ec5:
https://github.com/mapbox/mason/pull/144#issuecomment-205402583 is still accurate, so merging this would break the iOS SDK static library until we bump the minimum deployment target there.

kkaefer:
@1ec5 um, I don't think it is: https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/config.cmake#L2

1ec5:
That line determines the minimum deployment target of CMake-generated mbgl targets within ios.xcodeproj. But the iOS SDK itself has a [minimum deployment target of iOS 7.0](https://github.com/mapbox/mapbox-gl-native/blob/531a1c3070a098b5b31082c0500f0a3ce6a500bb/platform/ios/ios.xcodeproj/project.pbxproj#L2283).

Issue #252, opened 2016-09-22
Use clang++ toolchain without any runtime dependency on libstdc++


springmeyer:
Currently the clang++3.8 compiler we target on linux depends on `libstdc++-5-dev`. This is problematic for several reasons:

  - This means that packages whose `.travis.yml` needs custom apt deps, and therefore use the `addons: apt: packages` per matrix need to add `libstdc++-5-dev` to their packages otherwise that dep is not cascaded from the main `.travis.yml` and then the clang++ setup fails.
  - This means that packages that do not need c++14 or otherwise don't need to depend on the latest libstdc++ headers will still build against them. This is not ideal if those packages want to avoid using recent GLIBC and GLIBCXX symbols in order to support older libstdc++ versions. While this is not a primary consideration for mason, it may be important to some use cases to not force downstream users to depend on the latest libstdc++ symbols.

A solution I see for this problem is:

  - Change clang++ itself to link against its own libc++ instead of linking to the system libstdc++.
  - Then clang++ would not have a runtime dependency on libstdc++, making the `addon` override problem go away
  - Then packages needing libstdc++-5-dev or libstdc++-6-dev (for c++11 or c++14 support)  could opt-in as needed. How to do that is a separate issue to address.

/cc @kkaefer

springmeyer:
> Change clang++ itself to link against its own libc++ instead of linking to the system libstdc++.

This works! The upcoming llvm 3.8.1 and 3.9.0 packages at https://github.com/mapbox/mason/pull/253 now do this.

> Then clang++ would not have a runtime dependency on libstdc++, making the addon override problem go away

✅ 

> Then packages needing libstdc++-5-dev or libstdc++-6-dev (for c++11 or c++14 support) could opt-in as needed. How to do that is a separate issue to address.

So going to close this issue and will track how to selectively upgrade libstdc++ per package at https://github.com/mapbox/mason/issues/267

Issue #251, opened 2016-09-22
Ensuring consistent compiler version for linux binaries


springmeyer:
Mason should be more strict about using consistent compiler versions for all mason binaries.

This ticket describes some context of why this is important, background, proposed ideas for a solving this and related issues.

## Context 

We've been mixing clang 3.5 and 3.8 binaries for a while now in mason for Linux packages. Old packages are built with 3.5, and new ones published are built with 3.8. This is not okay going forward. Recently an issue arose in mapnik-vector-tile on Linux that indicates rare bugs can surface when clang++ 3.5 and 3.8 binaries are mixed: https://github.com/mapbox/mapnik-vector-tile/issues/216. In this case 1000s of unit tests were passing and only one was failing. But that failing test was only solvable by rebuilding all binaries to consistently use clang 3.8. So the tentative takehome is that mixing clang 3.5 and 3.8 binaries works 99% of the time. That is not good enough. For a binary deployment solution to be solid the binaries need to work 100% of the time.

### linux
Currently we are targeting clang 3.8. But 3.8.1 and 3.9.0 are now out and we should get a plan in place to upgrade to those compilers safely. Per the `Context` above we have evidence to suggest that mixing binaries between clang++ versions (at least 3.5 <->3.8) is not safe (see `Context` below for more info). At this point we don't have any evidence that mixing binaries between clang (any version) and gcc 5 is problematic (as long as `-D_GLIBCXX_USE_CXX11_ABI=0` is set when using gcc5 to ensure linking works).

### macOS and iOS
Currently most .travis.yml target `osx_image: xcode7.3`. We have no evidence to suggest there is are problems mixing binaries between xcode versions, but ideally we come up with a good mechanism to track the latest stable, which is now `osx_image: xcode8`

### Android
We use the cross compiler so this is already explicit and controlled, no action needed.

## Ideas for improving mason

- Add dev mode to mason core to enable easily rebuilding all packages against a new compiler without overwriting any existing binaries. Applications using Mason should be able to opt-in to the freshly compiled binaries when upgrading their mason version.
- For builds: add environment assertion that runs before packages are built that assert:
  - CXX and CC environment settings make sense
  - CXX --version and CC --version make sense
- Add post-build assertions that test for GLIBC and GLIBCXX requirements and ensure no > than expected  
- Vet every .travis.yml for consistency of format and toolchain bootstrap.

/cc @kkaefer @jfirebaugh @jakepruitt @tmpsantos

[...]

PR #253, opened 2016-09-23
Add llvm 3.8.1 and 3.9.0

springmeyer:
Adds packages for llvm 3.8.1, 3.9.0, and 4.x from git HEAD. These include binaries for clang, compiler-rt, libcxx, clang-tool-extras, etc. The goal is, after testing, to move Mason to defaulting to either 3.8.1 or 3.9.0 for its core linux compiler as per #251. Which version can be discussed at #251 - this PR is about getting new compilers available.

Adds a package for `clang++` which is a slimmed down repackaging of the llvm binaries for a given version to include just the pieces necessary to build C/C++ programs:
  - clang++/clang
  - lib/clang headers
  - share/
  - libLTO (for -flto)
  - libc++/libc++abi

Adds packages for `clang-tidy` and `clang-format`, which like the `clang++` package above, is a slimmed down binary repackaged from the source llvm build with the minimal command line tool for fast installs.

Note: there is no package named `clang` for `3.8.1` or `3.9.0`. What was previous named `clang` for the `3.8.0` is now called `llvm` because it includes a whole lot more than just `clang`. To get `clang/clang++` install the `clang++` package.

New in these packages:
 - clang++ now links against libc++ on linux to avoid libstdc++ runtime dep (#252)
 - now using md5 checks for llvm downloads
 - support for macOS 10.12

These packages were built using a very powerful AWS `Ubuntu 12.04.5 LTS (Precise Pangolin)` instance (not travis) because llvm takes too long to compile for travis. These packages were compiled with Mason provided clang 3.8.0.

Next actions:

 - [ ] @springmeyer is going to branch mason to test building new mason packages with these compilers
  - A custom s3 prefix or path will be used to avoid overwriting existing binaries while testing)
  - An s3 sync command will be run to bootstrap that new s3 path with a few key build dependencies: https://gist.github.com/springmeyer/b7f0351df61400c3175235872ddfc640
 - [ ] @springmeyer is going to branch mbgl-native, osrm, and mapnik to test these new compilers + newly compiled deps
   - Will test both `-D_GLIBCXX_USE_CXX11_ABI=0` and `-D_GLIBCXX_USE_CXX11_ABI=1` against gcc5/6
 - [x] review/questions on this PR: @kkaefer @jfirebaugh @tmpsantos @1ec5 
 - [x] Merge this PR, then move forward with #251
 - [ ] Make modifications to .travis.yml following #252 (no need for `libstdc++-5-dev` flag for C programs).

### DIFF
--- a/scripts/llvm/base/common.sh
+++ b/scripts/llvm/base/common.sh
@@ -51,6 +51,7 @@ function setup_release() {
     curl_get_and_uncompress "http://llvm.org/releases/${MASON_VERSION}/libunwind-${MASON_VERSION}.src.tar.xz"         ${MASON_BUILD_PATH}/projects/libunwind      
     curl_get_and_uncompress "http://llvm.org/releases/${MASON_VERSION}/lld-${MASON_VERSION}.src.tar.xz"               ${MASON_BUILD_PATH}/tools/lld               
     curl_get_and_uncompress "http://llvm.org/releases/${MASON_VERSION}/clang-tools-extra-${MASON_VERSION}.src.tar.xz" ${MASON_BUILD_PATH}/tools/clang/tools/extra 
+    curl_get_and_uncompress "http://llvm.org/releases/${MASON_VERSION}/lldb-${MASON_VERSION}.src.tar.xz"              ${MASON_BUILD_PATH}/tools/lldb
 }
 
 function mason_load_source {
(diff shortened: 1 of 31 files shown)


scripts/llvm/base/common.sh
```diff
@@ -5,6 +5,14 @@ MASON_LIB_FILE=bin/clang
 
 . ${MASON_DIR}/mason.sh
 
+function get_md5() {
+    if [[ $(uname -s) == 'Darwin' ]]; then
+        md5 -q $1
+    else
+        md5sum $1 | cut -d ' ' -f1
+    fi
```
kkaefer: Why use MD5 instead of a newer hash algo like `shasum`?

springmeyer:
Next action on this PR is to start testing https://github.com/mapbox/mason/issues/267 (now that the clang++ binaries no longer have a runtime dependency on libstdc++)

springmeyer:
Going to merge this to get the recent compiler versions in master mason. Then I'll hit the followup tasks at https://github.com/mapbox/mason/issues/267 and https://github.com/mapbox/mason/issues/251

PR #271, opened 2016-10-19
Only set MASON_DYNLIB_SUFFIX=tbd for system packages

springmeyer:
Apple started using tbd for system libraries. But the rest of the world still uses dylib. Mason does package a few libraries as shared (notably tbb) that need to know MASON_DYNLIB_SUFFIX is `dylib` on OSX.

refs https://github.com/mapbox/mason/pull/131#r56075159

/cc @kkaefer @jfirebaugh


springmeyer:
Will plan to merge this tomorrow unless there are objections.

PR #281, opened 2016-11-14
Rename TRAVIS_TOKEN to MASON_TRAVIS_TOKEN

springmeyer:
The `mason trigger` command needs access, locally, to the travis "Access token" generated via `travis token`. This renames the variable we expect to avoid clashing with the behavior of the travis ci tool (https://github.com/travis-ci/travis.rb/blob/b75c38bcdb736d1011e0d2a0022344a9870acaf4/README.md#environment-variables).

This ensures that `travis token --org` or `travis token --pro` will return the correct token from the travis API and not the token that might be set in a mason developers environment.

/cc @mapbox/cpp

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,9 +170,9 @@ This command only works if the package has already been installed. When run it s
 
 In order to ensure that all prebuilt binaries are consistent and reproducible, we perform the final build and publish operation on Travis CI.
 
-First set the `TRAVIS_TOKEN` environment variable. You can do this either by installing the `travis` gem and running `travis token` or by using `curl` to hit the Travis api directly. See details on this below.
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Why wouldn't you want them to be the same?

springmeyer:
Per chat with @jfirebaugh - there is no benefit from them being the same. When they are the same `travis token --pro` and `travis token --org` return the env variable value, which may be wrong and is totally confusing until you know `travis token` does nothing if `TRAVIS_TOKEN` is set. So, my strong opinion is that we should rename the variables to avoid this conflict causing any confusion in the future.

PR #294, opened 2016-12-02
Rebuild compile flags for Android

kkaefer:
The current CMake toolchain file in the Android NDK uses different compile flags compared to what we have [encoded in `mason.sh`](https://github.com/mapbox/mason/blob/master/mason.sh#L135). For now, we should update the flags in there. In the long term, compilers distributed via mason should have a `toolchain.sh` and `toolchain.cmake` file like the [GCC compilers already have](https://github.com/mapbox/mason/blob/master/scripts/gcc/5.3.0-cortex_a9/toolchain.cmake)

### DIFF
--- a/mason.sh
+++ b/mason.sh
@@ -132,17 +132,15 @@ elif [ ${MASON_PLATFORM} = 'android' ]; then
         *) export MASON_ANDROID_ABI=${MASON_ANDROID_ABI:-arm-v7}
     esac
 
-    CFLAGS="-fpic -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -fno-integrated-as -fomit-frame-pointer -fstrict-aliasing -Wno-invalid-command-line-argument -Wno-unused-command-line-argument"
-    LDFLAGS="-no-canonical-prefixes -Wl,--warn-shared-textrel -Wl,--fatal-warnings -Wl,-z,now -Wl,-z,relro"
+    CFLAGS="-g -DANDROID -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -Wa,--noexecstack -Wformat -Werror=format-security"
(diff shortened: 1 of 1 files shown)


kkaefer:
For reference, I "extracted" them from android.toolchain.cmake with the following `CMakeLists.txt` file:

```cmake
message("ANDROID_COMPILER_FLAGS: ${ANDROID_COMPILER_FLAGS}")
message("ANDROID_LINKER_FLAGS: ${ANDROID_LINKER_FLAGS}")
```

and ran

```bash
cmake .. \
    -DANDROID_ABI=armeabi \
    -DANDROID_NDK="${ANDROID_HOME}/ndk-bundle" \
    -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_TOOLCHAIN_FILE="${ANDROID_HOME}/ndk-bundle/build/cmake/android.toolchain.cmake" \
    -DANDROID_TOOLCHAIN=clang \
    -DANDROID_STL=c++_static
```

for all Android architectures, then manually extracted the common flags, and added the architecture-specific flags. The most significant change is that the NDK now seems to prefer Thumb  for ARM. This reduced overall code size by about 600 KB(!) for a build of Mapbox GL Native for Android/arm-v7.

kkaefer:
Once this is merged, we should rebuild all Android libraries with the new SDK and the new compile flags to take advantage of code size reductions in dependent libraries.

springmeyer:
> In the long term, compilers distributed via mason should have a toolchain.sh and toolchain.cmake file like the GCC compilers already have

@kkaefer Should I add this for the llvm/clang++ packages I maintain?

kkaefer:
Ideally yes, but we should probably create a mini-specification of what information we need in these files.

springmeyer:
@kkaefer okay, please ticket with more details when you think of it.

PR #297, opened 2016-12-02
Boost 1.62 fixes

springmeyer:
This fixes a few errors specific to the boost 1.62 packages. Previous packages were somewhat immune since they do not have mason.ini files. This fixes the generation of the mason.ini for boost 1.62.0 such that:

  - The header-only `boost` package does not report ldflags and static_libs
  - The library name is correct for ldflags for the actual libraries

### DIFF
--- a/scripts/boost/1.62.0/.travis.yml
+++ b/scripts/boost/1.62.0/.travis.yml
@@ -1,6 +1,9 @@
 language: generic
 
-sudo: false
+matrix:
+  include:
+    - os: linux
+      sudo: false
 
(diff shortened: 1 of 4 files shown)


springmeyer:
Will wait to merge this until these packages are published:

```
./mason build boost_libsystem 1.62.0
./mason build boost_libtest 1.62.0  
./mason build boost_libthread 1.62.0
```

springmeyer:
> Will wait to merge this until these packages are published:
done

Issue #213, opened 2016-07-21
PSA: Don't use moving targets as release names


kkaefer:
I noticed there are a few packages that use branch names, or something like `latest` as their release version. Please don't do this; mason doesn't automatically update those releases when the source code changes, so a `latest` package is outdated very quickly.

If you want to create a package that is not an actual release, please use a date and/or a commit hash as the version name, e.g. [`rapidjson 2016-07-20-369de87`](https://github.com/mapbox/mason/tree/master/scripts/rapidjson/2016-07-20-369de87). While the commit hash is enough for uniquely identifying the release, adding a date means that it's easier for humans to figure out what the newest release is.

/cc @springmeyer @jfirebaugh @mikemorris @brunoabinader @rclark @flippmoke @artemp @mourner @morganherlocker @tmpsantos @mapsam @yhahn @lbud @danpat @ansis @bleege @bhousel @TheMarex @1ec5 @daniel-j-h @friedbunny @ajashton @BergWerkGIS @camillacaros

springmeyer:
The currently remaining packages with `latest` are below:

- icu/latest
- mapnik/latest
- minjur/latest
- osmium-tool/latest
- tippecanoe/latest
- valgrind/latest

Of those `mapnik/latest` is the only one I know that is auto-updating so all the others should be transitioned to be packages pinned to the latest released version of a gitsha.

lightmare:
May I ask why the "moving target" build scripts were removed in response to this?

I understand why it's a bad idea to publish and/or depend on such package. That could've been prevented by removing their `.travis.yml` (or renaming to `no.travis.yml` to preserve that config), right?

What if I just want to `./mason build mapnik master` locally? I have to `git log --stat scripts/mapnik/` to figure out which of the hash-named directories is the most recent (1), copy and edit that. If there was a master/latest script, that would work out of the box, and you could also use it as the base for tagged version scripts (instead of what's the most recent hash).

(1) why noone followed @kkaefer's advice to use date with hash, or some other monotonic sequence, for example `git describe`

PR #298, opened 2016-12-06
update to c++11 for tippecanoe

camilleanne:
cc @springmeyer

### DIFF
--- a/scripts/tippecanoe/1.15.1/.travis.yml
+++ b/scripts/tippecanoe/1.15.1/.travis.yml
@@ -1,0 +1,21 @@
+language: cpp
+
+sudo: false
+
+matrix:
+  include:
+    - os: osx
+      compiler: clang
+    - os: linux
+      compiler: clang
+      addons:
+        apt:
+          sources: [ 'ubuntu-toolchain-r-test' ]
+          packages: [ 'libstdc++-5-dev' ]
+
+script:
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+- ./mason link ${MASON_NAME} ${MASON_VERSION}
+
+after_success:
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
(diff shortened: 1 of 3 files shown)


springmeyer:
@camilleanne - for your downstream need can the latest tagged release of tippecanoe be used? (1.15.1) /cc @ericfischer 

Asking because it is ideal to use 1.15.1 if that is sufficient. If not - if you need an unreleased version - then the convention now after https://github.com/mapbox/mason/issues/213 is to use the gitsha for versioning instead of `latest`.

springmeyer:
https://github.com/mapbox/mason/blob/master/scripts/luabind/e414c57bcb687bb3091b7c55bbff6947f052e46b/script.sh is an example of a C++ library that uses an unreleased gitsha, and is therefor named that way.

camilleanne:
@springmeyer yes, `1.15.1` will be fine for our needs -- should I expand this PR to pin to that version?

springmeyer:
> @springmeyer yes, 1.15.1 will be fine for our needs -- should I expand this PR to pin to that version?

Yes. I would recommend creating:

  - using this branch
  - create a new package at `scripts/tippecanoe/1.15.1/`
  - copying the script.sh and .travis.yml over from the `latest` folder
  - edit appropriately
  - commit, and trigger a new build for the new version
  - then we'll merge this (okay to leave the latest, we can cull later).

ericfischer:
Thanks for updating the mason scripts @camilleanne! I'll be glad to have this example to copy next time I tag a new release.

camilleanne:
@springmeyer ok, done and done.

springmeyer:
@camilleanne okay, I'm testing and fixing up a few things locally:

 - libprotobuf is no longer a dependency
 - upgrading the sqlite3 version used
 - using the release tarball: https://github.com/mapbox/tippecanoe/archive/${MASON_VERSION}.tar.gz

PR #299, opened 2016-12-07
Added wagyu 0.2.0 and removed 0.1.0

flippmoke:


springmeyer:
Because I see there is now a wagyu `0.3.0` I'll close this. Please re-open a new PR with the latest version (unless you think there is value to packaging both v0.2.0 and v0.3.0).

PR #302, opened 2016-12-15
Added geometry 0.9.0-gcc-4.9

brunoabinader:


springmeyer:
@brunoabinader should this be merged or should the work be upstreamed into geometry.hpp? /cc @jfirebaugh

Issue #310, opened 2017-01-04
Boost 1.63.0


springmeyer:
Add package for boost 1.63.0 and upgrade to latest cmake that supports it.

springmeyer:
boost 1.63.0 is now packaged.

daniel-j-h:
CMake 3.7.2 now contains the FindBoost module for Boost 1.63: https://github.com/Kitware/CMake/commit/a8b576244f729e543795750eb6c9e188c0e1fe66

PR #311, opened 2017-01-04
Add boost 1.63

springmeyer:
Refs #310


springmeyer:
Before merging these last packages needed triggered (hit rate limits so will pick this up on Monday):

```
./mason trigger boost_libtest 1.63.0
./mason trigger boost_libthread 1.63.0
```

PR #312, opened 2017-01-04
Add cmake 3.7.1

springmeyer:
Refs #310


daniel-j-h:
Unfortunately 3.7.1 does not yet support Boost 1.63

https://github.com/Kitware/CMake/blob/v3.7.1/Modules/FindBoost.cmake#L728-L746

PR #314, opened 2017-01-06
Added Linux + Cortex A9 and i686 build for Mapbox GL Native dependencies

tmpsantos:
Also fix some cross compilation issues on Linux.


.travis.yml
```diff
@@ -11,12 +11,16 @@ matrix:
...
   if [[ ${MASON_PLATFORM:-unset} == 'unset' ]] || [[ ${MASON_PLATFORM} == 'osx' ]] || [[ ${MASON_PLATFORM} == 'linux' ]]; then
-    echo "Installing clang for OSX/Linux via 'install' target"
-    source utils/toolchain.sh
-    which ${CC}
-    ${CC} -v
-    which ${CXX}
-    ${CXX} -v
+    # If platform is Linux or OSX, we might still be cross compiling if the
+    # MASON_PLATFORM_VERSION is not the same as the host platform.
+    if [[ ${MASON_PLATFORM_VERSION} == `uname -m` ]] || [[ -z ${MASON_PLATFORM_VERSION} ]]; then
```
springmeyer: This also needs to skip android and ios builds. So I think we need to keep the logic to only opt-in if `osx` or `linux` is the `MASON_PLATFORM` otherwise this will incorrectly trigger installing this clang++ version for those platforms where we want to use the platform specific compiler.

scripts/libpng/1.6.25/script.sh
```diff
@@ -5,20 +5,42 @@ MASON_VERSION=1.6.25
...
 
+function mason_prepare_compile {
+    if [[ ${MASON_PLATFORM_VERSION} == "cortex_a9" ]]; then
+        cd $(dirname ${MASON_ROOT})
+        ${MASON_DIR}/mason install zlib_shared ${ZLIB_SHARED_VERSION}
+        ${MASON_DIR}/mason link zlib_shared ${ZLIB_SHARED_VERSION}
+
+        MASON_ZLIB_CFLAGS="$(${MASON_DIR}/mason cflags zlib_shared ${ZLIB_SHARED_VERSION})"
+        MASON_ZLIB_LDFLAGS="-L$(${MASON_DIR}/mason prefix zlib_shared ${ZLIB_SHARED_VERSION})/lib"
+    fi
```
springmeyer: 👍  Can you add a brief comment about why this is needed for cortex? (for our future selves when we forget)

scripts/webp/0.5.1/script.sh
```diff
@@ -19,6 +19,11 @@ function mason_load_source {
 
 function mason_compile {
     export CFLAGS="${CFLAGS:-} -Os"
+
+    if [[ ${MASON_PLATFORM_VERSION} == "cortex_a9" ]]; then
+        CFLAGS="${CFLAGS:-} -mfloat-abi=softfp -mfpu=neon"
```
springmeyer: Should these flags be added for all `cortex_a9` packages or just webp?

scripts/zlib_shared/1.2.8/.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+
+sudo: false
+
+compiler: clang
+
+matrix:
+  include:
+    - os: linux
+      env: MASON_PLATFORM_VERSION=cortex_a9
+
```
springmeyer: Let's also build on linux x86_64 so that we have another platform that tests this package.

scripts/webp/0.5.1/script.sh
tmpsantos: Don't think so. This directly affects the floating point precision, so it should be evaluate case by case.

.travis.yml
tmpsantos: But when building for Cortex A9, `MASON_PLATFORM` will be set to `linux`, but `MASON_PLATFORM_VERSION` will be `cortex_a9`.

tmpsantos:
@springmeyer heads up that I'm now also adding the i686 build for these libraries.

.travis.yml
springmeyer: 👍

scripts/webp/0.5.1/script.sh
springmeyer: Okay, can you add a comment about why this is used for webp then?

PR #326, opened 2017-01-23
Bundles CMake 3.7.2 for Boost 1.63, see #310.

daniel-j-h:
We bundle Boost 1.63 but CMake before 3.7.2 lack `FindBoost.cmake` support for Boost 1.63.

cc @springmeyer


springmeyer:
Thanks. I'll land this today and it will be part of mason v0.6.0 (which I'll also tag later today)

PR #334, opened 2017-01-31
Nsis 2.51

springmeyer:


springmeyer:
Closing without merging. I found that `nsis 2.x` segfaulted upon use. I did not dig into why but rather moved to 3.x and will be using that per #335

PR #337, opened 2017-02-06
Add sdf-glyph-foundy mason target

ChrisLoer:
Adding mason script for https://github.com/mapbox/sdf-glyph-foundry

@springmeyer

### DIFF
--- a/scripts/sdf-glyph-foundry/0.1.0/.travis.yml
+++ b/scripts/sdf-glyph-foundry/0.1.0/.travis.yml
@@ -1,0 +1,8 @@
+language: cpp
+sudo: false
+
+script:
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+
+after_success:
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
--- a/scripts/sdf-glyph-foundry/0.1.0/.travis.yml
+++ b/scripts/sdf-glyph-foundry/0.1.0/.travis.yml
@@ -1,5 +1,4 @@
 language: generic
-sudo: false
 
 script:
 - ./mason build ${MASON_NAME} ${MASON_VERSION}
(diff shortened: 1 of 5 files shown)


scripts/sdf-glyph-foundry/0.1.0/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: cpp
```
springmeyer: `language: generic` works here and should allow the machines to boot a bit quicker. This is the reason I default to it instead of `cpp` which adds nothing for our usage.

scripts/sdf-glyph-foundry/0.1.0/script.sh
```diff
@@ -0,0 +1,33 @@
...
+}
+
+function mason_cflags {
+    echo "-I${MASON_PREFIX}/include"
+}
+
+function mason_ldflags {
+    :
+}
+
```
springmeyer: Also define a `mason_static_libs` that is a no-op just like `mason_ldflags`. This is minor, but helps avoid a `pkg-config` warning.

scripts/sdf-glyph-foundry/0.1.1/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: generic
+sudo: false
```
springmeyer: remove this `sudo:false` to avoid odd travis behavior of also publishing via `osx`. For header-only builds we need only one publish job

ChrisLoer: Done.... although I have very little understanding of what the change actually means.

springmeyer:
@ChrisLoer this looks ready to me. I've rebased to ensure it merges cleanly. Anything else you see needing done before merging?

PR #343, opened 2017-02-08
[WIP] tbb 2017_20161128 [skip ci]

springmeyer:
- Adds new tbb package from latest release.
- Adds static library builds

### DIFF
--- a/scripts/tbb/2017_20161128/.travis.yml
+++ b/scripts/tbb/2017_20161128/.travis.yml
@@ -1,0 +1,14 @@
+language: generic
+
+matrix:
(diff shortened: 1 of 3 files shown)


springmeyer:
This is ready to test /cc @daniel-j-h @danpat - You can pull this branch and then install like:

```bash
./mason install tbb 2017_20161128 
```

You should see both `.a` and `.so` at `ls $(./mason prefix tbb 2017_20161128)/lib/`.

daniel-j-h:
Thanks - integrating into osrm-backend :+1:

PR #344, opened 2017-02-08
Cross-compiler toolchain fixes

brunoabinader:
This PR:
- Fixes `MASON_LIB_FILE` for gcc 5.3.0-i686 - this is used in various places in `mason.sh` to perform package checks.
- Added no-op `mason_static_libs` for all gcc cross-compiler toolchain packages.

Fixes the build error in https://travis-ci.org/mapbox/mason/jobs/199379072#L168.

### DIFF
--- a/scripts/gcc/4.9.2-cortex_a9-hf/script.sh
+++ b/scripts/gcc/4.9.2-cortex_a9-hf/script.sh
@@ -28,4 +28,8 @@ function mason_ldflags {
     :
 }
 
+function mason_static_libs {
+    :
+}
+
 mason_run "$@"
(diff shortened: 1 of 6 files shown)


scripts/gcc/4.9.2-cortex_a9-hf/script.sh
```diff
@@ -28,4 +28,8 @@ function mason_ldflags {
     :
 }
 
+function mason_static_libs {
```
tmpsantos: Is this because when you want to link statically with the tool chain you do it via `-static` ?

brunoabinader: Yes - that's a sanity check we apply on all mason cross-compiler packages.

PR #350, opened 2017-02-09
Further minimize or-tools build

springmeyer:
This PR attempts to tame the # of dependencies needed to build `or-tools` and moves to pulling the remaining dependencies from mason packages instead of letting or-tools build them.

/cc @danpat

### DIFF
--- a/scripts/gflags/2.1.2/.travis.yml
+++ b/scripts/gflags/2.1.2/.travis.yml
@@ -1,0 +1,21 @@
+language: generic
+
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode8
+      compiler: clang
+    - os: linux
+      sudo: false
+      addons:
+        apt:
+          sources:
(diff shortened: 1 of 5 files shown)


springmeyer:
```
In file included from src/base/mutex.cc:14:
In file included from /usr/lib/gcc/x86_64-linux-gnu/4.6/../../../../include/c++/4.6/thread:38:
/usr/lib/gcc/x86_64-linux-gnu/4.6/../../../../include/c++/4.6/chrono:535:6: error: no matching constructor for initialization of 'duration' (aka 'std::chrono::duration<long, std::ratio<1, 1000000> >')
          : __d(__t.time_since_epoch())
            ^   ~~~~~~~~~~~~~~~~~~~~~~
```

Noting for my future self that ^^ should be fixed by https://github.com/mapbox/mason/pull/350/commits/6845ddc632798a588219c1a1599571ebec156443

PR #353, opened 2017-02-14
Remove tippecanoe 1.9.7

springmeyer:
@mattficke noticed the tippecanoe 1.9.7 package (added in #https://github.com/mapbox/mason/pull/143) was missing binaries. It looks like it never had binaries published so it never worked this way. This PR queues it up for removal in the next release. Existing users of mason that might be source compiling using the v1.9.7 package will continue working until they upgrade to mason > v0.7.0, when the package will be removed.

The latest tippecanoe is 1.16.3 and that is the recommend package to install.

/cc @ericfischer @rclark


springmeyer:
Closing, this will land in https://github.com/mapbox/mason/pull/362

PR #356, opened 2017-02-18
Error on undefined variables

springmeyer:
I'm surprised we've been living without this. Undefined variables can cause major headaches in package scripts.sh. I just hit this with a `cp -r ${VARIABLE}/ ${MASON_PREFIX}/bin` where VARIABLE was undefined so mason attempted to copy my whole user folder. #neveragain

### DIFF
--- a/mason
+++ b/mason
@@ -4,7 +4,7 @@ MASON_COMMAND=$1 ; shift
 MASON_NAME=$1 ; shift
 MASON_VERSION=$1 ; shift
 
-set -e
+set -eu
 set -o pipefail
 
 MASON_RELEASED_VERSION="0.7.0"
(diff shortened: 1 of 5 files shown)


springmeyer:
@brunoabinader - can you kick the tires on this and ensure the cross compiles you maintain still work? I had to make minor changes to `MASON_PLATFORM_VERSION` variable handling to avoid errors with this - would be good to confirm it still works. If you do find breakages then it will be a sign we are missing unit test coverage.

springmeyer:
This caught that `${MASON_SDK_PATH}` is undefined in the `zlib system` package at  https://github.com/mapbox/mason/blob/mapnik-3.0.13/scripts/zlib/system/script.sh#L26

PR #357, opened 2017-02-20
Remove gcc 4.9.2-{i686,cortex_a9,cortex_a9-hf}

brunoabinader:
These are no longer supported/maintained.


kkaefer:
Do we have a *need* to remove them? If not, we should just leave them.

brunoabinader:
> Do we have a need to remove them? 

They no longer work e.g. `MASON_PLATFORM=linux MASON_PLATFORM_VERSION=i686 mason install gcc 4.9.2` throws errors instead. This is due to the recent changes in mason cross-compiler support.

PR #359, opened 2017-02-22
Added cxx11abi version of gtest 1.8.0

tmpsantos:
Fix linkage errors.

### DIFF
--- a/scripts/gtest/1.8.0-cxx11abi/.travis.yml
+++ b/scripts/gtest/1.8.0-cxx11abi/.travis.yml
@@ -1,0 +1,19 @@
+language: generic
+
+matrix:
+  include:
+    - os: linux
+      sudo: true
+      addons:
+        apt:
+          sources:
+           - ubuntu-toolchain-r-test
+          packages:
+           - libstdc++-5-dev
+
+script:
+- sudo perl -i -p -e "s/# define _GLIBCXX_USE_DUAL_ABI 0/# define _GLIBCXX_USE_DUAL_ABI 1/g;" /usr/include/x86_64-linux-gnu/c++/5/bits/c++config.h
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+
+after_success:
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
--- a/scripts/gtest/1.8.0-cxx11abi/script.sh
+++ b/scripts/gtest/1.8.0-cxx11abi/script.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+LIB_VERSION=1.8.0
+
+MASON_NAME=gtest
+MASON_VERSION=${LIB_VERSION}-cxx11abi
+MASON_LIB_FILE=lib/libgtest.a
+
+. ${MASON_DIR}/mason.sh
+
+function mason_load_source {
+    mason_download \
+        https://github.com/google/googletest/archive/release-${LIB_VERSION}.tar.gz \
+        a40df33faad4a1fb308282148296ad7d0df4dd7a
+
+    mason_extract_tar_gz
+
+    export MASON_BUILD_PATH=${MASON_ROOT}/.build/googletest-release-${LIB_VERSION}/googletest
+}
+
+function mason_compile {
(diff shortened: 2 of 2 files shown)


tmpsantos:
@springmeyer I need your help here, I can't get to make this package to build proper cxx11abi binaries. It is building using g++5 and it is also doing the string replace trick but when I test it locally it fails to link. :(

springmeyer:
@tmpsantos you need to set `-D_GLIBCXX_USE_CXX11_ABI=1` somewhere in your script. You probably need something like:

```
export CXXFLAGS="${CXXFLAGS} -D_GLIBCXX_USE_CXX11_ABI=1"
export CFLAGS="${CFLAGS} -D_GLIBCXX_USE_CXX11_ABI=1"
```

If the program is both C and C++.

Also there is no reason you need to override the clang++ install. This ABI setting, as we found, is controlled not by the compiler but by the `_GLIBCXX_USE_DUAL_ABI` and `D_GLIBCXX_USE_CXX11_ABI` flags, in the libstdc++ headers.

scripts/gtest/1.8.0-cxx11abi/.travis.yml
```diff
@@ -0,0 +1,25 @@
+language: cpp
```
springmeyer: `language:generic` is fine, `cpp` adds nothing.

scripts/gtest/1.8.0-cxx11abi/.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+- sudo perl -i -p -e "s/# define _GLIBCXX_USE_DUAL_ABI 0/# define _GLIBCXX_USE_DUAL_ABI 1/g;" /usr/include/x86_64-linux-gnu/c++/5/bits/c++config.h
+- export CXX=g++-5
+- export CC=gcc-5
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+
+after_success:
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
+
+install:
+- true
```
springmeyer: please remove this and default to the clang++ compiler, per our other c++abi build (libprogram_options)

scripts/gtest/1.8.0-cxx11abi/script.sh
```diff
@@ -0,0 +1,50 @@
...
+    mason_extract_tar_gz
+
+    export MASON_BUILD_PATH=${MASON_ROOT}/.build/googletest-release-${LIB_VERSION}/googletest
+}
+
+function mason_build {
+    mason_load_source
+    mason_step "Building for Platform '${MASON_PLATFORM}/${MASON_PLATFORM_VERSION}'..."
+    cd "${MASON_BUILD_PATH}"
+
```
springmeyer: `mason_load_source` is called automatically, so these lines are unneeded. Where they left in for debugging purposes?

scripts/gtest/1.8.0-cxx11abi/script.sh
```diff
@@ -0,0 +1,50 @@
...
+    mkdir -p ${MASON_PREFIX}/include/gtest
+    cp -rv include ${MASON_PREFIX}
+
+    rm -rf build
+    mkdir -p build
+    cd build
+
+    cmake \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_INSTALL_PREFIX=${MASON_PREFIX} \
```
springmeyer: Add `-DCMAKE_CXX_FLAGS="${CXXFLAGS:-} -D_GLIBCXX_USE_CXX11_ABI=1` to compile for the c11_abi.

scripts/gtest/1.8.0-cxx11abi/script.sh
tmpsantos: It is only called automatically if you used `mason_compile` instead of `mason_build`. Not sure why the original package was using `build` but I will change that.

tmpsantos:
@springmeyer thanks for helping and reviewing, it worked now. I updated the patch.

springmeyer:
Great @tmpsantos - merging now.

PR #364, opened 2017-02-27
Add Shapeinfo library

oxidase:
Added " The Shapefile C Library provides the ability to write simple C programs for reading, writing and updating (to a limited extent) ESRI Shapefiles, and the associated attribute file (.dbf)."
http://shapelib.maptools.org/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
     - node v4.7.3
     - node v6.9.5
     - stxxl 1.4.1-1
+    - libshp2 1.3.0
 
  - Removed
     - tippecanoe 1.9.7 (never worked, never had binaries published)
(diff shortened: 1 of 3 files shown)


springmeyer:
Thanks @oxidase. The `script.sh` looks good. I will test it shortly to make sure it builds okay on OS X. Can you please replace the `.travis.yml` with https://github.com/mapbox/mason/blob/master/scripts/freetype/2.7.1/.travis.yml - that builds for just osx/linux in a way consistent with most recent packages (I presume we don't need shplib builds for OSRM for any platforms besides osx/linux).

springmeyer:
@oxidase I rebased to fix the conflict and added optimization flags to the build. We add `CFLAGS` environment per platform in mason's default environment, but this does not (yet) include `-O3`. It looks like this was overriding the `-O2` in the makefile, so I've fixed it now to pass `-O3`. I'll plan to merge this soon. Currently aws s3 is down which prevents proper testing. Will wait on that, then merge.

oxidase:
@springmeyer just added a patch for makefile: in the original makefile environments variables are ignored. i also removed exporting CFLAGS.

springmeyer:
Great, merging now. Thanks for getting this going.

PR #366, opened 2017-02-28
fix submodule example

mcwhittemore:
The example was referencing geometry.hpp rather than geometry

@flippmoke for the review


springmeyer:
Thanks for catching this @mcwhittemore @flippmoke - merging now.

PR #368, opened 2017-03-02
[WIP] Perf

springmeyer:
Perf binaries from latest linux kernel. Also adds dependencies:

 - liblzma.a / xz 5.2.3 (compression)
 - slang 2.3.1 (used for perf console UI)
 - elfutils 0.168

This package works on all recent ubuntu versions >= precise (and RHEL 6), independent of kernel version of the host.

`>= node 4.4.0` is needed for using `--perf_basic_prof_only_functions` to generate `/tmp/perf-<pid>.map` files for perf

### DIFF
--- a/scripts/elfutils/0.168/.travis.yml
+++ b/scripts/elfutils/0.168/.travis.yml
@@ -1,0 +1,21 @@
+language: generic
+
+matrix:
+  include:
+    - os: linux
+      sudo: false
+      addons:
+        apt:
+          sources:
+           - ubuntu-toolchain-r-test
+          packages:
+           - g++-4.9
+
+# note: elfutils must be compiled with gcc (see script.sh)
+install:
+  - export CXX=g++-4.9
+  - export CC=gcc-4.9
+
+script:
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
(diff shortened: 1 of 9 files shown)


daniel-j-h:
Sweet! What's your take on having a full set of profiling tools available in mason?

I can think of
- Brendan Gregg's [flamegraphs](http://www.brendangregg.com/flamegraphs.html)
- https://github.com/google/perf_data_converter
- https://github.com/google/pprof

maybe others. cc @oxidase

springmeyer:
@daniel-j-h yes, this is the first step towards my goal of easy tooling for flamegraphs. I've not used pprof, but interested to try it and I think packaging it is within scope.

oxidase:
@daniel-j-h i could add to your lists also [valgrind](http://valgrind.org/). It is a "swiss army knife" tool for analysis: callgrind for call graphs and performance analysis, cachegrind for cache optimization, massif for memory profiling, helgrind for threads interaction analysis.

springmeyer:
@oxidase we currently package v3.12.0: https://github.com/mapbox/mason/blob/master/scripts/valgrind/3.12.0/script.sh

springmeyer:
Before merging:

  - consider building perf on trusty to get more recent kernal
  - check if libeberty is needed for demangling c++
  - check if libnuma, libbfd, or libaudit are needed

springmeyer:
Keeping branch live for a week or so until I update scripts pointing at it.

PR #373, opened 2017-03-06
Ensure broken packages stop hard on `mason_config` error

### LINKED ISSUE #370: MASON_SYSROOT: unbound variable
Ticket to track some unexpected fallout from #356 due to latent bugs.

It looks like there is an always undefined `MASON_SYSROOT` variable used [here](https://github.com/mapbox/mason/blob/master/mason.sh#L577-L584) in the default `mason_cflags` added in https://github.com/mapbox/mason/commit/944a64b37148ee7a0405cdf85be45bdc1e6e9af3

This is now silently leading to the `mason.ini` to be not fully built out because the command quietly fails on travis: `/Users/travis/build/mapbox/mason/mason.sh: line 577: MASON_SYSROOT: unbound variable`

So, there are two bugs perhaps:

  - Need to get rid of the unbound variable
  - Need to ensure that `pkg_config` failures stop the build rather than continuing

springmeyer:
This changes `mason` such that if `mason_config` fails, we will stop abruptly with an error and will avoid potentially publishing packages with a busted `mason.ini`.

It also:

 - Fixes #370 to ensure that `mason.ini` can be correctly generated even when `MASON_SYSROOT` is not passed (to avoid a regression from #356)
 - Fixes a few packages to not error due to invalid usage uncovered by #356

### DIFF
--- a/mason.sh
+++ b/mason.sh
@@ -565,7 +565,7 @@ function mason_try_binary {
 
 function mason_pkgconfig {
     MASON_PKGCONFIG_FILES=""
-    for pkgconfig_file in ${MASON_PKGCONFIG_FILE}; do
+    for pkgconfig_file in ${MASON_PKGCONFIG_FILE:-}; do
         MASON_PKGCONFIG_FILES="${MASON_PKGCONFIG_FILES} ${MASON_PREFIX}/${pkgconfig_file}"
     done
     echo pkg-config ${MASON_PKGCONFIG_FILES}
--- a/mason.sh
+++ b/mason.sh
@@ -505,7 +505,8 @@ function mason_config {
 function mason_write_config {
     local INI_FILE
     INI_FILE="${MASON_PREFIX}/mason.ini"
-    echo "`mason_config`" > "${INI_FILE}"
+    MASON_INI_DATA=$(set -e;mason_config)
+    echo "${MASON_INI_DATA}" > "${INI_FILE}"
     mason_substep "Wrote configuration file ${INI_FILE}:"
     cat ${INI_FILE}
 }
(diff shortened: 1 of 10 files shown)


mason.sh
```diff
@@ -505,7 +505,8 @@ function mason_config {
 function mason_write_config {
     local INI_FILE
     INI_FILE="${MASON_PREFIX}/mason.ini"
-    echo "`mason_config`" > "${INI_FILE}"
+    MASON_INI_DATA=$(set -e;mason_config)
+    echo "${MASON_INI_DATA}" > "${INI_FILE}"
```
kkaefer: Any reason they're now split up into two lines?

mason.sh
```diff
@@ -564,24 +565,32 @@ function mason_try_binary {
...
-    for pkgconfig_file in ${MASON_PKGCONFIG_FILE}; do
+    for pkgconfig_file in ${MASON_PKGCONFIG_FILE:-}; do
         MASON_PKGCONFIG_FILES="${MASON_PKGCONFIG_FILES} ${MASON_PREFIX}/${pkgconfig_file}"
     done
     echo pkg-config ${MASON_PKGCONFIG_FILES}
 }
 
 function mason_cflags {
+    if [[ ! ${MASON_PKGCONFIG_FILE:-} ]]; then
+        mason_error " The MASON_PKGCONFIG_FILE variable not found in script.sh. Please either provide this variable or override both mason_cflags"
```
kkaefer: message seems to end abruptly

mason.sh
springmeyer: If line 508 fails, now the `set- e` will prevent any blank mason.ini from being saved. In local testing it seemed this way needed.

mason.sh
springmeyer: Fixed in 493b3b3

springmeyer:
Merging this: the known outcome is that some older packages will fail to rebuild. This is desirable to avoid those packages ever being rebuilt (or new packages created as copies) with broken `mason.ini`. /cc @brunoabinader @tmpsantos @flippmoke @jfirebaugh

PR #374, opened 2017-03-07
Add strict check for xcrun output

### LINKED ISSUE #210: Catch broken xcode setup?
We should likely catch the case where `xcrun --sdk macosx --show-sdk-path` returns empty and throw a startup error.

Refs https://github.com/Project-OSRM/node-osrm/issues/211#issuecomment-232298224

/cc @1ec5 @jfirebaugh

springmeyer:
Fixes #210

/cc @MoKob @1ec5


1ec5:
Fixed a typo in the error message in 5e39d3bbcd88a48431586ce151dea3c26c872be3.

Issue #377, opened 2017-03-07
Deprecate "mason link" command


springmeyer:
The `mason link` command collapses all linked packages into an unversioned directory structure. This was inspired by the hombrew `link` command and is useful if you want to install lots of dependencies with mason and point to a single, predictable, location for their headers and libs. 

However, per chat with @kkaefer we want to move to a future where all apps using mason for deps consistently use the mason client (either mason.cmake or mason bash client) to fetch versioned paths. Removing `mason link` will remove surface area to maintain and limit the amount of ways downstream apps use mason.  By using the `mason client` we can help downstream apps using mason packages without suffering through difficultings of using existing `*-config` tools that likely will not work with the relative structure of mason packages, like `pkg-config`, `freetype-config`, or `gdal-config`. For example code like https://github.com/mapbox/mason/blob/1319207a0e88eb842156ca027876be3770aca587/scripts/gdal/2.1.3/script.sh#L154-L174 could be avoided if we started to make no promise that `gdal-config` works in mason packages (or even removed packaging it) because we only support downstream apps using the `mason` client to get include and lib paths.

Note: there are a lot of uses of `mason link`, so only proposing deprecating and not removing. I imagine I'll need several months at least to migrate all usage off `mason link` in external repos like  
 - https://github.com/mapbox/carmen-cache/blob/master/binding.gyp
 - https://github.com/mapbox/wagyu/blob/92d8ef5974316f60c33556c611d18691659bd87e/Makefile#L16-L25
 - and internal packages:

```
$ grep 'mason link' scripts/ -r
scripts//gdal/1.11.1-big-pants/.travis.yml:- ./mason link ${MASON_NAME} ${MASON_VERSION}
scripts//gdal/2.0.2/.travis.yml:- ./mason link ${MASON_NAME} ${MASON_VERSION}
scripts//gdal/dev/.travis.yml:- ./mason link ${MASON_NAME} ${MASON_VERSION}
scripts//gdal/dev/script.sh:    ${MASON_DIR}/mason link $1 $2
scripts//geos/3.4.2/.travis.yml:- ./mason link ${MASON_NAME} ${MASON_VERSION}
scripts//geos/3.5.0/.travis.yml:- ./mason link ${MASON_NAME} ${MASON_VERSION}
scripts//geowave-jace/0.8.7/script.sh:    ${MASON_DIR}/mason link $1 $2
scripts//gflags/2.1.2/script.sh:    ${MASON_DIR}/mason link cmake 3.7.2
[...]

PR #380, opened 2017-03-07
libosmium 2.12.0 and osmium-tool 1.6.0 

springmeyer:
Added latest libosmium and osmium-tools

Before merging:

  - [x] ~~remove use of `mason link` in these packages - refs #377~~ Too much work, staying with using `link`
  - [x] publish binaries

/cc @joto @freenerd @ingalls @mattficke


ingalls:
Rad! Thanks for getting this out @springmeyer

PR #381, opened 2017-03-08
harfbuzz 1.4.4-ft

springmeyer:
/cc @artemp


artemp:
👍 looks good. It fails to download `expat` on travis though ^.

Issue #385, opened 2017-03-13
llvm 4.0.0 package


springmeyer:
The official llvm 4.0.0 release is out, so next steps are to:

 - [x] pin the existing llvm 4.0.0 mason package to the release tarballs (currently pointed to git head)
 - [x] rebuild and republish 4.0.0
 - [ ] create a 5.0.0 package that points at upstream llvm to begin allowing that to be tested

I'm out today, but will plan to hit this tomorrow.

/cc @kkaefer @daniel-j-h

kkaefer:
> create a 5.0.0 package that points at upstream llvm
Can we please name it `5.0.0-pre-<gitsha>`? or something like that?

springmeyer:
> Can we please name it 5.0.0-pre-<gitsha>? or something like that?

👍 @kkaefer

springmeyer:
Per #386 llvm 4.0.0 packages are now available. Not yet found the time to build 5.0.0-pre packages.

springmeyer:
> create a 5.0.0 package that points at upstream llvm to begin allowing that to be tested

Closing, not going to move on this now. Will reconsider if there is a specific need.

PR #386, opened 2017-03-14
llvm 4.0.0

springmeyer:
Refs #385


springmeyer:
Packages are ready and published. They were done locally on OS X and on a large ubuntu precise ami for linux (since travis is too slow to do the builds). Methods are documented at https://github.com/mapbox/mason/blob/llvm-4.0.0/scripts/llvm/base/README.md#details-of-builds.

PR #394, opened 2017-03-22
Fixed unbounded vars when not compiling png for cortex_a9 or i686

edkimmel:
If you aren't compiling png for cortex_a9 or i686,  the MASON_ZLIB_* variables are never set, and cause an error. This fixes that case.


springmeyer:
Excellent, thanks for catching and fixing this. Will merge now.

PR #268, opened 2016-10-14
libboost libprogram_options 1.62.0 cxx11abi

tmpsantos:
Fixes build on recent Linux distros.

PR #397, opened 2017-03-31
geojson 0.4.1

ivovandongen:
Small update to geojson to include https://github.com/mapbox/geojson-cpp/issues/31

### DIFF
--- a/scripts/geojson/0.4.1-cxx11abi/.travis.yml
+++ b/scripts/geojson/0.4.1-cxx11abi/.travis.yml
@@ -43,5 +43,7 @@ matrix:
       env: MASON_PLATFORM=android MASON_ANDROID_ABI=mips-64
 
 script:
+# enable DUAL_ABI per https://github.com/mapbox/mason/issues/319
+- sudo perl -i -p -e "s/# define _GLIBCXX_USE_DUAL_ABI 0/# define _GLIBCXX_USE_DUAL_ABI 1/g;" /usr/include/x86_64-linux-gnu/c++/5/bits/c++config.h
 - ./mason build ${MASON_NAME} ${MASON_VERSION}
 - ./mason publish ${MASON_NAME} ${MASON_VERSION}
--- a/scripts/geojson/0.4.1-cxx11abi/.travis.yml
+++ b/scripts/geojson/0.4.1-cxx11abi/.travis.yml
@@ -19,13 +19,16 @@ matrix:
       env: MASON_PLATFORM=osx
       compiler: clang
     - os: linux
-      sudo: false
+      sudo: true
       addons:
         apt:
           sources:
            - ubuntu-toolchain-r-test
           packages:
            - libstdc++-5-dev
+      before_script:
+        # enable DUAL_ABI per https://github.com/mapbox/mason/issues/319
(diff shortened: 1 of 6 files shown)


springmeyer:
Please hold on this. It looks like this package is intended to be built using the CXX11 ABI but I can tell that it is not actually ending up working. @ivovandongen I realize you are just here after releasing a new geojson-cpp release that needs packaged. I'm happy to fix this up so that it actually builds a CXX11 ABI version per what @tmpsantos and I learned over at https://github.com/mapbox/mason/pull/268#issuecomment-272380280.

springmeyer:
Also more context: my understanding is that because we previously did not know how to built CXX11 ABI packages in mbgl we are using geojson-cpp as a header only library. So I'll make some changes to this PR that fixes to build to publish a working CXX11ABI package. Then we could revisit using geojson-cpp as a precompiled library in mbgl, if @mapbox/gl desires. We should have both options I think.

springmeyer:
The above commits:

  - Fix the cxx11abi package to actually build against that ABI
  - Renamed the package with the `-cxxabi` modifier (currently the default is ABI 0 and packages with ABI 1 are named with `-cxxabi` until a future breaking mason version that might switch that default)
  - Created a non cxxabi version
  - Now compiling with clang++ like all other packages (I presume the G++ attempt there was in the hope that would enable the cxxabi. But in fact it is controlled by the libstdc++ headers and not the compiler)

I noticed one other issue. This package is depending on https://github.com/mapbox/geojson-cpp/blob/b7470e2049b2866e516134abe201d88ab2a69893/Makefile#L4-L6 for the dependency versions. These are currently the same as mbgl https://github.com/mapbox/mapbox-gl-native/blob/master/CMakeLists.txt#L41-L54 but could spell trouble if they drift (/cc @kkaefer). So I think a better approach would be to let the mason script control the versions rather that the build of geojson-cpp. This could require build modifications so those can be overriden. I can take a look at this next week.

springmeyer:
Also noticed that there is not a tag yet for 0.4.1 at https://github.com/mapbox/geojson-cpp/releases. @ivovandongen let me know if you plan to do that?

kkaefer:
> I noticed one other issue. This package is depending on https://github.com/mapbox/geojson-cpp/blob/b7470e2049b2866e516134abe201d88ab2a69893/Makefile#L4-L6 for the dependency versions. These are currently the same as mbgl https://github.com/mapbox/mapbox-gl-native/blob/master/CMakeLists.txt#L41-L54 but could spell trouble if they drift (/cc @kkaefer)

To circumvent this issue, we're using this library as [`HEADER_ONLY`](https://github.com/mapbox/mapbox-gl-native/blob/74ff4354623a644eb3c293ec818cebbf9c45077c/CMakeLists.txt#L52) and [compile it in the project](https://github.com/mapbox/mapbox-gl-native/blob/74ff4354623a644eb3c293ec818cebbf9c45077c/src/mbgl/util/geojson.cpp) so that it uses the same dependency headers.

ivovandongen:
@springmeyer Thanks for fixing this up. I've added the tag here: https://github.com/mapbox/geojson-cpp/releases/tag/v0.4.1

springmeyer:
Thanks @ivovandongen for the tag.

Thanks @kkaefer for the confirmation that mbgl will not be impacted by drifting versions currently. 

This that the cxx11abi package built as a library (and included in this PR) is not needed by mbgl. However I think we should keep this package going since it may be needed in the future by mbgl or another application. To avoid getting in trouble with drift, in the future mason should offer ways to assert that dependency versions are not in conflict - refs #403

PR #399, opened 2017-03-31
Added wagyu 0.4.2

flippmoke:
/cc @springmeyer for review and merge

### DIFF
--- a/scripts/wagyu/0.4.2/.travis.yml
+++ b/scripts/wagyu/0.4.2/.travis.yml
@@ -1,0 +1,8 @@
+language: cpp
+sudo: false
+
+script:
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+
+after_success:
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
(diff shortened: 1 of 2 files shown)


scripts/wagyu/0.4.2/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: cpp
```
springmeyer: Technically this can be `language: generic`, which would boot faster. That is not going to matter, but it is what I prefer.

scripts/wagyu/0.4.2/.travis.yml
```diff
@@ -0,0 +1,8 @@
...
+sudo: false
+
+script:
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+
+after_success:
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
```
springmeyer: This is just fine, and a lot of existing mason files look like this. It is on my list however to do a big fixing pass. For your reference I'll share how this could be improved. It is a good idea to put the `publish` step inside the `script`. The reason is that otherwise it might look like the job passed but the publish could fail if network failed. If you put the publish inside script then the travis job will fail, helping you notice the problem and the need to send another `triggered` build.

springmeyer:
👍  made a few comments, but they are non-critical. Looks good to merge.

PR #402, opened 2017-04-04
Add thelink2012/any @ 8fef1e9 [skip ci]

jfirebaugh:
https://github.com/thelink2012/any

C++11-compatible implementation of [`std::experimental::any`](http://en.cppreference.com/w/cpp/experimental/any). Experimenting with use for https://github.com/mapbox/mapbox-gl-native/issues/7375.

### DIFF
--- a/scripts/any/8fef1e9/.travis.yml
+++ b/scripts/any/8fef1e9/.travis.yml
@@ -1,0 +1,8 @@
+language: generic
+sudo: false
+
+script:
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+
+after_success:
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
--- a/scripts/any/8fef1e9/.travis.yml
+++ b/scripts/any/8fef1e9/.travis.yml
@@ -1,5 +1,9 @@
 language: generic
-sudo: false
+
+matrix:
+  include:
+    - os: linux
+      sudo: false
 
 script:
 - ./mason build ${MASON_NAME} ${MASON_VERSION}
--- a/scripts/any/8fef1e9/script.sh
+++ b/scripts/any/8fef1e9/script.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+MASON_NAME=any
+MASON_VERSION=8fef1e9
+MASON_HEADER_ONLY=true
+
+GIT_HASH="8fef1e93710a0edf8d7658999e284a1142c4c020"
+
+. ${MASON_DIR}/mason.sh
+
+function mason_load_source {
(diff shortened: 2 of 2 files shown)


scripts/any/8fef1e9/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: generic
+sudo: false
```
springmeyer: Remove this `sudo:false`. Otherwise the inheritance from the main .travis.yml will confuse travis and it ends up building for OS X as well: https://travis-ci.org/mapbox/mason/builds/218571809.

I'm a little fuzzy on why this happens but my sense is it is because the main `.travis.yml` has `sudo:false` inside a matrix. And I think the global one here conflicts with the one inside the matrix at https://github.com/mapbox/mason/blob/d8860aa211c9d41f0fddc9892d29c5b4c9588a32/.travis.yml#L9. Not sure if this is just the way yaml works or a travis bug. But staying away from globals inside the package specific `.travis.yml` will avoid the problem.

springmeyer:
@jfirebaugh looks great. Very interesting library. One question. I noticed a few `#define` inside `any.hpp` that impact behavior: `ANY_IMPL_FAST_TYPE_INFO_COMPARE` (https://github.com/thelink2012/any/blob/master/any.hpp#L309) and `ANY_IMPL_ANYCAST_MOVEABLE` (https://github.com/thelink2012/any/blob/8fef1e93710a0edf8d7658999e284a1142c4c020/any.hpp#L401). If you know these should be used then adding them to the `mason_cflags` output would be a good idea. If you don't know then of course its not necessary. But I wanted to flag in case you had ideas.

scripts/any/8fef1e9/.travis.yml
springmeyer: I was fuzzy. Turns out that to avoid jobs being automatically added for both osx and linux we need to define a matrix in the packages' travis.yml like:

```yml
matrix:
  include:
    - os: linux
      sudo: false
```

jfirebaugh:
The preprocessor symbols are documented in [the README](https://github.com/thelink2012/any#defines), and based on that I don't think that we want to define either of them. No `ANY_IMPL_FAST_TYPE_INFO_COMPARE` because I don't know what I'm doing, and no `ANY_IMPL_ANY_CAST_MOVEABLE` so that we have a standards-compliant implementation.

scripts/any/8fef1e9/.travis.yml
jfirebaugh: 👍 https://travis-ci.org/mapbox/mason/builds/218903851

springmeyer:
👍

Issue #396, opened 2017-03-30
Lightweight install client in nodejs


springmeyer:
@kkaefer has proposed a lightweight bash client in https://github.com/mapbox/mason/tree/client which has the promise to be the standard way we install packages via the command line soon. I'm creating this ticket to log the need for another client that is portable to windows.

The problem is:

   - node-mapnik currently depends on header-only packages via npm (protozero and mapnik-vector-tile)
   - It was decided at https://github.com/mapbox/cpp/issues/20, that we want to move away from publishing header only libs to npm, and only do it to mason
   - this means that node-mapnik needs a mason client that runs on windows

springmeyer:
I think this client should be written in nodejs. The design would be:

  - separate nodejs module (does not need to be stored in mason repo)
  - published to npm
  - a node js c++ addon could depend on it as a devDep
  - it would be able to download a mason package and put it inside `mason_packages` in the same format as the bash mason client and the cmake mason client

springmeyer:
Closing, this: work is now underway on a node.js client at https://github.com/mapbox/mason-js

PR #407, opened 2017-04-12
Removes unbound var in or-tools script

daniel-j-h:
Downstream builds are failing because of bash unbound var errors.

Not sure when exactly this changed, all I could find in the changelog is

```
Fixed
    Unbound variables in sparsehash, bzip2, and harfbuzz packages
```

Can we automatically check all packages?

cc @springmeyer

### DIFF
--- a/scripts/or-tools/5.1/script.sh
+++ b/scripts/or-tools/5.1/script.sh
@@ -88,6 +88,5 @@ function mason_static_libs {
 }
 
 
-echo $DIR
 
 mason_run "$@"


springmeyer:
> Not sure when exactly this changed, all I could find in the changelog is

This started throwing in the v0.8.0 release per `mason now run with set -u to catch undefined variables in scripts` (https://github.com/mapbox/mason/pull/356) https://github.com/mapbox/mason/blob/master/CHANGELOG.md#080.

> Can we automatically check all packages?

Kicked off this work at https://github.com/mapbox/mason/pull/408

springmeyer:
> Downstream builds are failing because of bash unbound var errors.

Note: in the future we plan to finish the lightweight client (https://github.com/mapbox/mason/issues/396), such that all downstream installers of packages will only depend on the `mason.ini` and therefore be isolated from any breakages in the `script.sh` which will only be used for builds.

springmeyer:
Thanks for finding and fixing this bug @daniel-j-h - will merge now. This will go out in the upcoming v0.10.0 release.

PR #412, opened 2017-05-03
sparsepp

ghoshkaj:
Adding sparsepp  to mason


springmeyer:
Thanks @ghoshkaj. Per chat this is now published (`./mason trigger ...`) and this PR looks good. So I will merge. This will land as part of the v0.10.0 release. In the future it will be possible to install packages without upgrading the mason version - follow https://github.com/mapbox/mason/issues/396 for details.

PR #425, opened 2017-06-06
Mapnik 3.0.14

artemp:


scripts/mapnik/6e02595/.travis.yml
```diff
@@ -0,0 +1,19 @@
+language: generic
```
flippmoke: @artemp do we want to drop this version/folder?

springmeyer:
@artemp great start. What is needed to get this ready is:

  - Rebase this against latest master so that this branch is up to date with master
  - Remove the `6e02595` folder - let's focus just on packaging v3.0.14.
  - For the `3.0.14` package:
     - [ ] Remove these lines (https://github.com/mapbox/mason/blob/90748e99dbb36a53dce4357f499b99a753734342/scripts/mapnik/3.0.14/script.sh#L18-L22). They are redundant - for a formal release we want to pull from the tagged tarball (which is done by the `mason_download` logic above). The git clone is only needed for dev packages of `master`.
     - [ ] Copy this logic (https://github.com/mapbox/mason/blob/44aa9039f0a493a38cabf046567387a28328aceb/scripts/mapnik/3.0.13-1/script.sh#L70-L76) from the 3.0.13-1 package,
 which fixes a bug I originally had in the 3.0.13 package but fixed in the 3.0.13-1 package.
 The fix ensures that CXXFLAGS/LDFLAGS from mason are passed to mapnik.

scripts/mapnik/6e02595/.travis.yml
springmeyer: 👍  yes. Noted at https://github.com/mapbox/mason/pull/425#issuecomment-306500900

springmeyer:
Changes look good @artemp - One note: in the future please also squash your commits down to one or two key commits. In this case the history of the whole branch landed, which is less ideal because it includes additions and removals of code that never landed in master - which could be confusing when looking at the history.

PR #435, opened 2017-06-30
If mason is executed through a symlink, resolve the link

ericfischer:


springmeyer:
Landed as part of #437

Issue #438, opened 2017-07-05
Package LLVM 4.0.1 bugfix release


daniel-j-h:
- http://lists.llvm.org/pipermail/llvm-announce/2017-July/000074.html
- http://releases.llvm.org/download.html

springmeyer:
👍  I will take this on.

springmeyer:
Packages are published and new docs on the process to create them are ready for review at https://github.com/mapbox/mason/pull/452. Will close this now and track at #452

PR #452, opened 2017-07-26
LLVM 4.0.1

springmeyer:
## Context

The llvm package contains a nearly complete toolchain for building and linking C++ code. Normally this would be provided by the operating system (g++ toolchain on linux or XCode on mac). Instead we package the llvm suite of compilers in mason to enable upgrading your systems toolchain quickly via mason. The advantages of this are:

 - We can fully control the compiler version on OS X and Linux, and upgrade seamlessly
 - We can install extra development tools like lldb, clang-tidy, clang-format, and include-what-you-use 
 - We can package these extra development tools in a way that allows just the minimal tools to be installed, in a fast and lightweight way. For example, while the entire llvm toolchain is over 1.6GB, clang-tidy is just 21 MB and clang++ is just 44 MB. Packaging these tools separately allows them to be installed fast on travis.ci, circle, or locally.

## This PR

Adds llvm 4.0.1 and related sub-packages, closing #438. This improves upon #451, by additionally adding:

 - A readme describing exactly how to create a new llvm package
 - Updated script to automatically create packages for all the sub-packages (#451 missed include-what-you-use, lldb, and llvm-cov)
 - A fix to avoid building libc++ on OS X (since we don't use it/link to it and instead use the system version)
 - Adds a dockerfile to support consistent method of building on linux binaries
 - Modifies the `llvm/base/common.sh` to allow pointing at a custom gitsha for a given sub-package - needed in the case of include-what-you-use which did not have a tag that worked yet with llvm 4.x.


springmeyer:
@daniel-j-h tagged you for review on this since I know you've been interested in the process of creating these packages being documented. The docs are now at https://github.com/mapbox/mason/blob/d6dbdad607b4a61ce4b70a119aec4147db3e262b/scripts/llvm/base/README.md. Would appreciate your 👍  before merging.

scripts/llvm/base/README.md
```diff
@@ -1,40 +1,203 @@
...
+docker build -t mason-llvm -f utils/Dockerfile.llvm .
+```
+
+B. Run the docker image
+
+We run the docker image to build the package on linux. We map volumes such that the binary will end up on our host machine (to avoid needing to pass publishing credentials to docker).
+
+LLVM_VERSION="4.0.2"
+docker run -it --volume $(pwd):/home/travis/build/mapbox/mason mason-llvm \
+  ./mason build llvm ${LLVM_VERSION} && ./utils/llvm.sh build ${LLVM_VERSION}
```
daniel-j-h: I think this misses the code-blocks backticks

utils/Dockerfile.llvm
```diff
@@ -0,0 +1,25 @@
...
+
+COPY mason mason
+COPY mason.sh mason.sh
+COPY scripts scripts
+COPY utils utils
+
+RUN apt-get update -y && \
+ apt-get install -y vim python build-essential bash curl git-core ca-certificates software-properties-common python-software-properties --no-install-recommends
+
+RUN add-apt-repository -y ppa:ubuntu-toolchain-r/test && \
```
daniel-j-h: Hm I guess pull this ppa in for a recent C++ stdlib version. But why exactly? Does building llvm or any of the binaries need a recent stdlib? We do not link to the stdlib in the binaries anyway, no? I don't quite follow here.

scripts/llvm/base/README.md
springmeyer: 👍 added in c5596d5

utils/Dockerfile.llvm
springmeyer: Nice catch! This is only needed for `libstdc++6`: an updated version of that is needed at runtime by `cmake`. It is not a compile/link dependency of llvm, so I've added comments about this to the Dockerfile to make this clear: 1c73eb6.

PR #458, opened 2017-08-01
add catch 1.9.6

### LINKED ISSUE #454: Document header-only publishing
It'd be great to get a statement on using the `MASON_HEADER_ONLY=true` variable for publishing header-only libraries. Right now the only reference is in [mason.cmake](https://github.com/mapbox/mason#masoncmake). Perhaps adding it to the [build scripts variables section](https://github.com/mapbox/mason#variables) would suffice?

cc @springmeyer

### LINKED ISSUE #455: Publish "catch.hpp"
It'd be great to get the excellent library, Catch, up on Mason to avoid vendorizing it.

https://github.com/philsquared/Catch

I can pick this up!

cc @springmeyer

mapsam:
### What changed?

* adds [catch.hpp](https://github.com/philsquared/Catch) to resolve #455 
* resolves #454 by adding the `MASON_HEADER_ONLY`
* updates the script.sh docs with a note of how to generate the `git hash-object` locally

cc @springmeyer

### DIFF
--- a/scripts/catch/1.9.6/.travis.yml
+++ b/scripts/catch/1.9.6/.travis.yml
@@ -1,0 +1,19 @@
+language: generic
+
+sudo: false
+
+matrix:
+  include:
+    - os: osx
+      compiler: clang
+      osx_image: xcode8.2
+    - os: linux
+      compiler: clang
+      addons:
+        apt:
+          sources: [ 'ubuntu-toolchain-r-test' ]
+          packages: [ 'libstdc++-5-dev' ]
+
+script:
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
--- a/scripts/catch/1.9.6/script.sh
+++ b/scripts/catch/1.9.6/script.sh
@@ -1,0 +1,31 @@
+MASON_NAME=catch
+MASON_VERSION=1.9.6
+MASON_HEADER_ONLY=true
+
+. ${MASON_DIR}/mason.sh
+
+function mason_load_source {
+    mason_download \
+        https://github.com/philsquared/Catch/archive/v${MASON_VERSION}.tar.gz \
+        c8bf11d32c73c864ebf6b57e2a324a86ea39df82
+
+    mason_extract_tar_gz
+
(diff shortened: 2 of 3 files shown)


scripts/catch/1.9.6/.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+    - os: osx
+      compiler: clang
+      osx_image: xcode8.2
+    - os: linux
+      compiler: clang
+      addons:
+        apt:
+          sources: [ 'ubuntu-toolchain-r-test' ]
+          packages: [ 'libstdc++-5-dev' ]
+
```
springmeyer: Because header-only packages are platform agnostic, this needs only one matrix entry. (otherwise it will double-publish to the same location, which is harmless but not ideal). See https://github.com/mapbox/mason/blob/master/scripts/geometry/0.9.2/.travis.yml for how this should look.

mapsam: Woops, grabbed this from tippecanoe accidentally. Good catch! Updated in https://github.com/mapbox/mason/pull/458/commits/def305dee4ac30bc037a6631e999b773baefa6f1

scripts/catch/1.9.6/script.sh
```diff
@@ -0,0 +1,31 @@
...
+
+    mason_extract_tar_gz
+
+    export MASON_BUILD_PATH=${MASON_ROOT}/.build/catch-${MASON_VERSION}
+}
+
+# nothing to build, just copying single include header file
+function mason_compile {
+    mkdir -p ${MASON_PREFIX}/include/
+    cp -r single_include ${MASON_PREFIX}/include
```
springmeyer: I think it would be clearer to do `cp -r single_include/catch.hpp ${MASON_PREFIX}/include/`. That way if the layout ever changed, we'd know since this would break. e.g. I like hardcoding that is explicit here for what gets installed. Since the expectation is that this is a single header, we can copy just the single header...

mapsam: Done in https://github.com/mapbox/mason/pull/458/commits/73d8c0a25c79a099cb28f5214ff3a777d7a2bb26

mapsam:
Header files have been published @springmeyer - merging!

springmeyer:
👍  please feel free to merge this then

Issue #450, opened 2017-07-19
Mason for `webp` on iOS does not download nor build


roblabs:
Attempting to include `webp` in an iOS project via Mason.  There is no prebuilt package for iOS, so it attempts to build then fails at a step called `lipo`.

* I understand that iOS is not supported with a download package, 
* Any ideas why `lipo` fails?  _cf._ #400 

## Using these steps to reproduce

``` bash
mkdir ./mason
curl -sSfL https://github.com/mapbox/mason/archive/v0.13.0.tar.gz | tar --gunzip --extract --strip-components=1 --exclude="*md" --exclude="test*" --directory=./mason

export MASON_PLATFORM=ios
./mason/mason build webp 0.6.0
```

I have Xcode Version 8.3.3 (8E3004b) installed 

```
> clang -v
Apple LLVM version 8.1.0 (clang-802.0.42)
Target: x86_64-apple-darwin16.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

## Attempts to download, but not available
```
* Downloading binary package ios-8.0/webp/0.6.0.tar.gz...
* Binary not available yet for https://mason-binaries.s3.amazonaws.com/ios-8.0/webp/0.6.0.tar.gz
* Building for Platform 'ios/8.0'...

```

## Builds several flavors, but then fails at lipo 
```
input file lib-ios-arm64/libwebp.a is not a fat file
Non-fat file: lib-ios-arm64/libwebp.a is architecture: i386
* Creating Universal Binary...
fatal error: 
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: 
lib-isim-i386/libwebp.a and lib-isim-x86_64/libwebp.a have the same architectures (i386) 
and can't be in the same fat output file
```

![webp](https://user-images.githubusercontent.com/118112/28351689-401efdf2-6c06-11e7-8be9-f743057f6c11.png)

roblabs:
Until I can figure out mason for `webp`, I sorted out a solution for iOS using `WebP.framework`, https://github.com/roblabs/mapbox-gl-native#mapbox-gl--webp-for-ios

romanroibu:
[...]

PR #465, opened 2017-08-15
Fix webp build for iOS compilation

kkaefer:
Fixes https://github.com/mapbox/mason/issues/450

We were missing a `make clean`, so while we were configuring the project for the next architecture, we weren't actually building the files since they were already in place from a different build from the previous architecture.


kkaefer:
https://travis-ci.org/mapbox/mason/builds/264770223 has the builds for various platforms, including iOS.

springmeyer:
@kkaefer 👍

PR #468, opened 2017-08-23
Upgrade Tippecanoe to version 1.22.2

ericfischer:


scripts/tippecanoe/1.22.2/.travis.yml
```diff
@@ -0,0 +1,19 @@
+language: generic
+
+sudo: false
```
springmeyer: this should be nested inside the `matrix:include` items. Otherwise it will trigger an odd travis behavior that results in multiple builds.

Also `tippecanoe` does not technically require `libstdc++-5-dev` so I would recommend copying https://github.com/mapbox/mason/blob/master/scripts/cmake/3.8.2/.travis.yml to replace this file before you trigger builds.

ericfischer: Thanks @springmeyer. I'm not sure which package I copied the previous tippecanoe version's `.travis.yml` from, since it doesn't seem to quite match any of the others.

I don't think I have the permissions to run `mason trigger` so I would appreciate your help.

springmeyer: @ericfischer okay. To be able to run `mason trigger` follow the steps at `1. Using curl and travis api to generate MASON_TRAVIS_TOKEN` at https://github.com/mapbox/mason#trigger.

PR #488, opened 2017-10-19
Add earcut 0.12.4

brunoabinader:


scripts/earcut/0.12.4/.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: generic
+
```
springmeyer: Please add:

```
matrix:
  include:
    - os: linux
      sudo: false
```

That will prevent the (currently harmless, but nevertheless non-ideal) double publish on travis which I presume is happening currently). Note: we only recently realized this was happening and fixed in the `geometry.hpp` package, hence why the latest package creation docs recommend copying from geometry as a starting point:  https://github.com/mapbox/mason#header-only-package

brunoabinader: Sure :+1: but isn't the double CI caused by both the PR trigger and the manual `mason publish` I've made?

brunoabinader: Ah, nevermind - I think I got what you meant now :100:

PR #489, opened 2017-10-20
Various style fixes in mason.sh

WesleyAC:
I noticed a few style sort of issues in `mason.sh`, and fixed a few of them:

* Change backtick substitution to  `$(...)`
* Add doublequotes around some variables to prevent globbing + splitting
* Add `-r` flag to `read`
* Change a use of `-a` for boolean and to `&&`, since it's easier to reason about

It might make sense to run some sort of linting tool as part of your test suite to catch things like this. There are many different scripts available to do this, and I've definitely caught real bugs when using them in the past.

I don't have the test suite fully working on my computer (even on master), but all of these changes are fairly safe and I think Travis should build it when I submit this PR.

### DIFF
--- a/mason.sh
+++ b/mason.sh
@@ -2,14 +2,14 @@ set -eu
 set -o pipefail
 # set -x
 
-export MASON_ROOT=${MASON_ROOT:-`pwd`/mason_packages}
+export MASON_ROOT=${MASON_ROOT:-$(pwd)/mason_packages}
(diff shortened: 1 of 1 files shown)


springmeyer:
Thank you for this PR @WesleyAC - looks great, and I'm grateful for the fixes. Just rebased and will merge once green. And yes, we should definitely run the code through something like https://www.shellcheck.net. I've ticketed that idea at https://github.com/mapbox/mason/issues/549

PR #492, opened 2017-10-23
gdal-2.1.3: Add a default value for CXX variable.

tpikonen:
At least on my system the 'CXX' environment variable is not set by default. This makes the gdal-2.1.3 mason build fail. This PR adds a reasonable default value.

### DIFF
--- a/mason.sh
+++ b/mason.sh
@@ -114,6 +114,8 @@ elif [ ${MASON_PLATFORM} = 'linux' ]; then
     export CFLAGS="-fPIC"
     export LDFLAGS=""
     export CXXFLAGS="${CFLAGS} -std=c++11"
+    export CXX="${CXX:-c++}"
+    export CC="${CC:-cc}"
 
     if [ $(uname -m) != ${MASON_PLATFORM_VERSION} ] ; then
         # Install the cross compiler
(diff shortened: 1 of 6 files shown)


tpikonen:
There are actually many more scripts which assume CXX to be set without exporting it (e.g. mapnik, gdal). It's indeed best to make sure reasonable defaults are set in mason.sh, as they are for at least osx and android.

tpikonen:
I changed this PR to set the CC and CXX vars in mason.sh.

springmeyer:
How about something like:

```diff
+    export CXX="${CXX:-c++}"
+    export CC="${CC:-cc}"
```

That will allow them to be customized (which is important since we build with clang++ on travis by setting CXX). And it will use the more generic symlinks to the compilers (which I presume will work on more linux systems and bsd systems).

tpikonen:
Good idea, I've updated the PR.

tpikonen:
On second thought, setting CC=cc etc. obviously breaks boost build scripts as they are now. Using readlink and some heuristics in boost scripts probably can fix this.

tpikonen:
I added a commit to this PR, which determines the BOOST_TOOLSET* vars in boost build scripts from the CC variable. It supports only gcc and clang though.

springmeyer:
Going to close this since #692 landed. Please re-open new issues for anything no yet fixed.

PR #493, opened 2017-10-23
Do not assume PKG_CONFIG_PATH to be defined.

tpikonen:
There are some instances in mason scripts (mostly harfbuzz scripts) where PKG_CONFIG_PATH is extended and exported. This patch fixes the extension when PKG_CONFIG_PATH is empty.


springmeyer:
Thank you for catching and fixing this.

PR #494, opened 2017-10-23
ragel: Add -std=gnu++98 to CXXFLAGS.

tpikonen:
On my (linux, IA32) system mason build of ragel-6.9 fails with -Wnarrowing, due to use of -std=c++11. This patch adds -std=gnu++98 to CXXFLAGS, which is enough to make the compiler not to turn this warning into an error.

There is probably a cleaner way to fix this though.

### DIFF
--- a/scripts/ragel/6.9/script.sh
+++ b/scripts/ragel/6.9/script.sh
@@ -17,6 +17,7 @@ function mason_load_source {
 }
 
 function mason_compile {
+    export CXXFLAGS="${CXXFLAGS//-std=c++11} -std=c++03"
     ./configure --prefix=${MASON_PREFIX} \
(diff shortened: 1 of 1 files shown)


springmeyer:
The problem is that mason hardcodes `std=c++11`. So I recommend trying out this fix to remove it:

```
export CXXFLAGS="${CXXFLAGS//-std=c++11}"
```

Does that alone work or do you really also need `-std=gnu++98`. I presume ` -std=gnu++98` does not need to be added.

tpikonen:
No, just removing `-std=c++11` does not work. This does:

`export CXXFLAGS="${CXXFLAGS//-std=c++11} -std=c++03"`

(or using `-std=gnu++98` as in the original PR).

PR #499, opened 2017-10-25
Set BOOST_TOOLSET* from CC and CXX env variables.

tpikonen:
Build scripts for boost currently have the toolset hardcoded to clang. Most linux systems have gcc as the default compiler, which makes the build fail without installing an additional compiler. This patch sets the BOOST_TOOLSET* variables from the CC and CXX environment variables, which should be set to reasonable defaults in mason.sh.

### DIFF
--- a/scripts/boost/1.61.0/base.sh
+++ b/scripts/boost/1.61.0/base.sh
@@ -4,7 +4,7 @@
 
 export MASON_VERSION=1.61.0
 export BOOST_VERSION=${MASON_VERSION//./_}
-export BOOST_TOOLSET="clang"
-export BOOST_TOOLSET_CXX="clang++"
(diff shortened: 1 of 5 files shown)


springmeyer:
Thanks for noticing and fixing. Also thanks for all your other PRs - very appreciated. I am busy but I'll do my best to respond and merge when I can.

As far as this one: did you have a chance to test? I worry that these arguments expect a compiler name rather than a path and CXX/CC might provide a full path - can you test that? Then, happy to merge.

tpikonen:
You're correct, Boost build system does not work with BOOST_TOOLSET* set to binaries with full path. Fortunately the basename of the full paths works for (at least) the common compilers of gcc and clang. I've updated the patch accordingly.

PR #522, opened 2017-11-28
jq 1.5 build script

ingalls:
Another dep I use quite a bit that I'm hoping to get managed through our mason install list.

### DIFF
--- a/scripts/jq/1.5-239278fd/.travis.yml
+++ b/scripts/jq/1.5-239278fd/.travis.yml
@@ -1,0 +1,18 @@
+language: generic
+
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode8
+      compiler: clang
+    - os: linux
+      compiler: clang
+      sudo: false
+      addons:
+        apt:
+          sources: [ 'ubuntu-toolchain-r-test' ]
+          packages: [ 'libstdc++-4.9-dev', 'libssl-dev' ]
+
+script:
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
--- a/scripts/jq/1.5-239278fd/.travis.yml
+++ b/scripts/jq/1.5-239278fd/.travis.yml
@@ -8,6 +8,7 @@ matrix:
     - os: linux
       compiler: clang
       sudo: false
+      dist: trusty # for more recent automake
       addons:
         apt:
           sources: [ 'ubuntu-toolchain-r-test' ]
--- a/scripts/jq/1.5-239278fd/.travis.yml
+++ b/scripts/jq/1.5-239278fd/.travis.yml
@@ -11,7 +11,7 @@ matrix:
       addons:
         apt:
           sources: [ 'ubuntu-toolchain-r-test' ]
-          packages: [ 'libstdc++-4.9-dev', 'libssl-dev' ]
+          packages: [ 'libstdc++-4.9-dev' ]
 
 script:
 - ./mason build ${MASON_NAME} ${MASON_VERSION}
--- a/scripts/jq/1.5-239278fd/script.sh
+++ b/scripts/jq/1.5-239278fd/script.sh
@@ -9,7 +9,9 @@ MASON_LIB_FILE=bin/jq
 function mason_load_source {
     export MASON_BUILD_PATH=${MASON_ROOT}/.build/${MASON_NAME}-${MASON_VERSION}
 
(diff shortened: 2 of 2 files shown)


ingalls:
JQ Build > https://travis-ci.org/mapbox/mason/builds/308487511

scripts/jq/1.5-239278fd/.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+    - os: osx
+      osx_image: xcode8
+      compiler: clang
+    - os: linux
+      compiler: clang
+      sudo: false
+      addons:
+        apt:
+          sources: [ 'ubuntu-toolchain-r-test' ]
+          packages: [ 'libstdc++-4.9-dev', 'libssl-dev' ]
```
springmeyer: Minor, but I think you can remove `libssl-dev` here, right? That was only needed for wget and not jq, correct?

scripts/jq/1.5-239278fd/script.sh
```diff
@@ -0,0 +1,45 @@
...
+. ${MASON_DIR}/mason.sh
+
+function mason_load_source {
+    export MASON_BUILD_PATH=${MASON_ROOT}/.build/${MASON_NAME}-${MASON_VERSION}
+
+    git clone git://github.com/stedolan/${MASON_NAME}.git ${MASON_BUILD_PATH}
+    cd ${MASON_BUILD_PATH}
+    git checkout ${MASON_VERSION}
+
+    cd ../
```
springmeyer: Direct git clone works - nice job figuring this out - however I'd prefer to keep the `mason_download` usage when possible, for consistency. In this case I think you can follow the lead of https://github.com/mapbox/mason/blob/master/scripts/protozero/ccf6c39/script.sh which is an example of pulling from a specific gitsha while still using `mason_download`.

ooooo, now I'm seeing the `submodule update --init`. That must be why you are forced to git clone instead of pulling from a tag? If that is the case then ignore my above comment. Though I will say that here is another example of pulling using git: https://github.com/mapbox/mason/blob/master/scripts/mapnik/f02a25901/script.sh#L10-L14. The only real substantial difference is pulling via `https` rather than `git`, but wanted to provide that example nonetheless

scripts/jq/1.5-239278fd/script.sh
```diff
@@ -0,0 +1,45 @@
...
+function mason_compile {
+    git submodule update --init
+    autoreconf -fi
+
+    ./configure \
+        --prefix ${MASON_PREFIX} \
+        --with-oniguruma=builtin \
+        --disable-maintainer-mode
+
+    make -j${MASON_CONCURRENCY} LDFLAGS=-all-static
```
springmeyer: Please do `LDFLAGS="<your custom flag>" ${LDFLAGS:-}` instead. This is important since Mason on some platforms may be customizing LDFLAGS and we want to respect the mason provided environment values as well.

springmeyer:
> JQ Build > https://travis-ci.org/mapbox/mason/builds/308487511

@ingalls in the future please link this build instead: https://travis-ci.org/mapbox/mason/builds/308527889. This is the one (notice the lego icon) that represents the `./mason trigger wget ...` build and which actually tests the building and publishing of the package. The other one is just running the mason unit tests, which are general and should have not changed in this branch. In the future we plan to fix this so that branches only test the package build, but this is a todo still.

scripts/jq/1.5-239278fd/script.sh
ingalls: 👍 Yeah I had used mason download tied to the commit originally but then discovered I needed the submodules

ingalls:
Tracking triggered build here: (Lego blocks and all!) https://travis-ci.org/mapbox/mason/builds/309636094

ingalls:
@springmeyer Looks like I hit the green, this look good to merge?

springmeyer:
@ingalls you need to run `./mason trigger jq 1.5-239278fd` again and ensure that specific travis job goes green (and since OS X is slow, it it will take a wait).

springmeyer:
This is not yet working (looks like the linux build needs a newer version of autotools): https://travis-ci.org/mapbox/mason/builds/338278969

springmeyer:
Error is: `configure.ac:6: require Automake 1.14, but have 1.11.3` which should be fixed by 
524b9ff.

@ingalls can you run `./mason trigger jq 1.5-239278fd` and see if it works now (travis fully green)?

springmeyer:
Published and working in https://travis-ci.org/mapbox/mason/builds/338516875

PR #532, opened 2018-01-04
Add a <experimental/optional> polyfill

tmpsantos:
Useful for platforms like Windows that don't ship with it yet.


tmpsantos:
And yes, Mason header-only + CMake + Windows works fine.

PR #537, opened 2018-01-16
Enable custom CPU archs

flippmoke:
Completely untested and spur of the moment
/cc @kkaefer @springmeyer


springmeyer:
Squeezing performance by using native arch feels, to me, like a fun experiment. But I recommend not attempting to do this with mason, since its not designed to do this and doing it properly would require a significant lift. Based on talking with @flippmoke I think what you are after should be possible to implement for the new modular repos that don't depend on mason for anything except header only libs. So, perhaps ticket at places like vtcomposite or vtquery this idea if you still think it is worth while? Should be easy to implement in those repos without changes to mason.

PR #545, opened 2018-02-05
Advances to LLVM + profiling packages

springmeyer:
This improves the LLVM package:

 - now providing sanitized `libc++` binaries at `${PREFIX}/<sanitizer>` for:
   - `msan`: memory sanitizer (MemoryWithOrigins): `-fsanitize=memory -fsanitize-memory-track-origins`
   - `tsan`: thread sanitizer: `-fsanitize=thread`
   - `asan`: address sanitizer: `-fsanitize=address,undefined -fno-sanitize=vptr,function -fno-sanitize-recover=all`
 - now building `lldb` against `ncurses` and `libedit` for auto-completion support on linux
 - now consistently building `libc++` on both linux and osx (still not installing `libc++` on osx)
 - better incremental `llvm` builds using `ccache` (to aid in debugging local package changes). With this change re-builds take ~20 min in docker on mac, previously would take > 2 hours.
 - disables building https://github.com/include-what-you-use/include-what-you-use by default inside llvm package - this has become too brittle since official `iwyu` releases have not kept up with llvm releases

Also this:

 - Adds a package for the https://github.com/iovisor/bcc linux profiling tools
 - Fixes libedit to work on ubuntu precise (allows lldb build inside llvm to work on ubuntu precise)

/cc for visibility @mapbox/core-tech @daniel-j-h @kkaefer @karenzshea @GretaCB


springmeyer:
Adding that the sanitized libc++ builds are now available in the mason llvm packages of 5.0.0, 5.0.1, and 7.0.0. But I've found that it is easiest/ideal to use 7.0.0 due to the support for `-nostdlib++`. The instructions follow for linux:

First create a `test.cpp`:

```c++
#include <iostream>

int main() {
  std::clog << "hello\n";
  return 0;
}
```

Then install llvm 7.0.0 (in the future we'll create sub packages for libc++ to avoid needing all of llvm):

```
LLVM_VERSION="7.0.0"
./mason install llvm ${LLVM_VERSION}
BASE=$(./mason prefix llvm ${LLVM_VERSION})
SAN="msan"
```

Then build the test program:

```
${BASE}/bin/clang++ \
    -std=c++14 \
    -stdlib=libc++ \
    -nostdlib++ -nostdinc++ \
    -I ${BASE}/${SAN}/include/c++/v1 \
    -fsanitize=memory \
    test.cpp -o test-command \
    -Wl,--start-group \
    ${BASE}/${SAN}/lib/libc++abi.a \
    ${BASE}/lib/libunwind.a \
    ${BASE}/${SAN}/lib/libc++.a \
    -rtlib=compiler-rt
```

Then you should be able to do:

```
$ ./test-command 
hello
```

Linking should now show no sign of libcstdc++ (which indicates correct static linking to the sanitized libc++.a):

```
ldd test-command 
	linux-vdso.so.1 =>  (0x00007fff5b9fe000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ff4ef807000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007ff4ef5e9000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007ff4ef3e0000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007ff4ef1dc000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff4eee13000)
	/lib64/ld-linux-x86-64.so.2 (0x00007ff4efb22000)
```

Note: Big thanks to @nico who landed `-nostdlib++` in https://github.com/llvm-mirror/clang/commit/37cdc82da7d49a9fe3991eca89f44cd05d86fc55. If we did not have  `-nostdlib++` available, then we'd have needed to pass `-nostdlib`. This would have triggered needing to hand form up links to c libs and friends to avoid many many linker errors (since `-nostdlib` leads clang++ to avoid linking to both the c++ lib, the c lib, and other pieces needed for the c libs):

```
    -lc -ldl -lrt -lpthread -lm \
    /usr/lib/gcc/x86_64-linux-gnu/4.6.3/crtbegin.o \
    /usr/lib/gcc/x86_64-linux-gnu/4.6.3/crtend.o  \
    /usr/lib/x86_64-linux-gnu/crt1.o \
    /usr/lib/x86_64-linux-gnu/crti.o \
    /usr/lib/x86_64-linux-gnu/crtn.o
```

Also related: this early experiment in using custom libc++ on OS X: https://gist.github.com/springmeyer/4e89a4b24c35dcd1dcd8

PR #546, opened 2018-02-05
Libnghttp2 1.26.0

ivovandongen:
Adds libnghttp2 1.26.0

### DIFF
--- a/scripts/libnghttp2/1.26.0/.travis.yml
+++ b/scripts/libnghttp2/1.26.0/.travis.yml
@@ -1,0 +1,36 @@
+language: generic
+
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode8.2
+      compiler: clang
+    - os: linux
+      env: MASON_PLATFORM=linux
+      compiler: clang
+      sudo: false
+    - os: linux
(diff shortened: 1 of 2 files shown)


scripts/libnghttp2/1.26.0/script.sh
```diff
@@ -0,0 +1,38 @@
...
+MASON_VERSION=1.26.0
+MASON_LIB_FILE=lib/libnghttp2.a
+MASON_PKGCONFIG_FILE=lib/pkgconfig/libnghttp2.pc
+
+. ${MASON_DIR}/mason.sh
+
+
+function mason_load_source {
+    mason_download \
+        https://github.com/nghttp2/nghttp2/releases/download/v1.26.0/nghttp2-1.26.0.tar.gz \
```
springmeyer: Minor: please use `${MASON_VERSION}` in place of these hardcoded version numbers.

scripts/libnghttp2/1.26.0/.travis.yml
```diff
@@ -0,0 +1,41 @@
+language: generic
+
+matrix:
```
springmeyer: Missing `include:` so this matrix is not going to work right.

PR #548, opened 2018-02-06
android-ndk r16b

ivovandongen:
- Adds android-ndk r16b
- Uses r16b as the default


mason.sh
```diff
@@ -134,12 +134,12 @@ elif [ ${MASON_PLATFORM} = 'linux' ]; then
 
 elif [ ${MASON_PLATFORM} = 'android' ]; then
     case "${MASON_PLATFORM_VERSION:-}" in
-        arm-v5-9) export MASON_ANDROID_ABI=arm-v5 ;;
-        arm-v7-9) export MASON_ANDROID_ABI=arm-v7 ;;
+        arm-v5-14) export MASON_ANDROID_ABI=arm-v5 ;;
+        arm-v7-14) export MASON_ANDROID_ABI=arm-v7 ;;
```
kkaefer: Note that this might break build systems that explicitly rely on these platform versions

ivovandongen: 14 is the minimum for this version of the sdk. What do you want to do? Anything specific I could test?

kkaefer: https://github.com/search?q=user%3Amapbox+arm-v7-9&type=Code&utf8=✓

kkaefer: Elaborating on that: Changing this number prefix means that Mason clients will look in the `mason_packages/android-arm-v7-14/` folder instead of the `-9` folder both locally, and on the server. Since all of our existing packages are published with ABI level 9, this would be a significant departure, and we'd have to rebuild all Android packages for this higher ABI level. I believe this is something we should do, but it requires some time to get them all rebuilt before we can merge this to master.

Apart from changing `mason.sh`, we also need to update the [CMake client](https://github.com/mapbox/mason/blob/master/mason.cmake) and [Node.js clients](https://github.com/mapbox/mason-js).

ivovandongen: As discussed in chat, I changed the CMake client and will trigger new builds for the relevant packages after merge so that the -14 versions are available for downstream projects.

PR #559, opened 2018-02-21
tippecanoe 1.27.7 + libc++

springmeyer:
This adds the latest tippecanoe release in a unique and forward looking way. This switches from building against old libstdc++-4.9 (for some portability to >= ubuntu trusty) to statically linking the latest version of libc++, which enables better performance and fuller portability. These binaries should work against any linux system based on glibc (amazon linux, all debian systems, fedora/rhel, etc).

This uses the new llvm compiler package with supports static linking to libc++: https://github.com/mapbox/mason/pull/545#issuecomment-367082479

/cc @mick @ericfischer

### DIFF
--- a/scripts/tippecanoe/1.27.7-libstdcxx/.travis.yml
+++ b/scripts/tippecanoe/1.27.7-libstdcxx/.travis.yml
@@ -1,0 +1,19 @@
+language: generic
+
+sudo: false
+
+matrix:
+  include:
+    - os: osx
+      compiler: clang
+      osx_image: xcode8.2      
+    - os: linux
+      compiler: clang
+      addons:
+        apt:
+          sources: [ 'ubuntu-toolchain-r-test' ]
+          packages: [ 'libstdc++-5-dev' ]
+
+script:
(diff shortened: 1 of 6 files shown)


mapsam:
@springmeyer would you like to include this in the 0.18.0 release? Happy to merge it in once the builds have been triggered.

springmeyer:
@mapsam thanks for asking. Before merging this these packages need be benchmarked on linux to ensure that `tippecanoe 1.27.7` is as fast or faster than the `tippecanoe 1.27.7-libstdcxx` package. Once that is confirmed I'll plan to drop the `tippecanoe 1.27.7-libstdcxx` package and move ahead with the current `tippecanoe 1.27.7` being linked to libc++ going forward and being the default package. This will mean that dockerfiles installing this tippecanoe package will no longer need to `apt-get install libstdc++-5-dev` and therefore will also no longer need to `apt-get install -y software-properties-common python-software-properties` or `add-apt-repository -y ppa:ubuntu-toolchain-r/test` to be enable the PPA to upgrade libstdc++.

millzpaugh:
Test are 🍏 !

PR #568, opened 2018-03-07
Added iOS & Android builds for protobuf 3.4.1

brunoabinader:


springmeyer:
Looks good to have 👍 . I see that the builds are failing with

```
google/protobuf/stubs/common.cc:142: error: undefined reference to '__android_log_write'
google/protobuf/stubs/common.cc:150: error: undefined reference to '__android_log_write'
```

Looks like you'll need to provide this workaround: https://github.com/google/protobuf/issues/2719

I recommend patching by generating a diff file and dynamically applying it during the `mason_compile` step. Details on this approach at https://github.com/mapbox/mason/blob/master/contributing.md#creating-new-packages

brunoabinader:
@springmeyer saw your message after pushing this change: https://github.com/mapbox/mason/pull/568/commits/b13bf07dd55d09226e7472d940cb3feb10118863

Would that suffice? Otherwise I can generate the diff :+1:

springmeyer:
> @springmeyer saw your message after pushing this change: b13bf07

👍 that is even better than I proposed. Great that you should be able to fix without diffing.

PR #580, opened 2018-03-15
openfst 1.6.7

boblannon:
this PR adds an update to the openfst script, targeting the most recent version.

### DIFF
--- a/scripts/openfst/1.6.7/.travis.yml
+++ b/scripts/openfst/1.6.7/.travis.yml
@@ -1,0 +1,15 @@
+language: generic
+
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode8.2
+      compiler: clang
+    - os: linux
+      env: MASON_PLATFORM=linux
+      compiler: clang
+      sudo: false
+
+script:
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
--- a/scripts/openfst/1.6.7/script.sh
+++ b/scripts/openfst/1.6.7/script.sh
@@ -17,7 +17,7 @@ function mason_load_source {
 }
 
 function mason_compile {
-    export CFLAGS="${CFLAGS:-} -std=c++11 -O3 -DNDEBUG"
+    export CFLAGS="${CFLAGS:-} -O3 -DNDEBUG"
     export CXXFLAGS="${CXXFLAGS:-} -std=c++11 -O3 -DNDEBUG"
     if [[ $(uname -s) == 'Darwin' ]]; then
         export LDFLAGS="${LDFLAGS:-} -stdlib=libc++"
--- a/scripts/openfst/1.6.7/script.sh
+++ b/scripts/openfst/1.6.7/script.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+
+MASON_NAME=openfst
+MASON_VERSION=1.6.7
+MASON_LIB_FILE=lib/libfst.a
+
+. ${MASON_DIR}/mason.sh
+
+function mason_load_source {
+    mason_download \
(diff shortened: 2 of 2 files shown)


scripts/openfst/1.6.7/.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: generic
+
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode8.2
+      compiler: clang
+    - os: linux
+      env: MASON_PLATFORM=linux
```
springmeyer: No harm having `env: MASON_PLATFORM=linux` but it is also redundant/unneeded since this is the default.

boblannon:
thanks! will do.

boblannon:
the linux build is failing.  this error is raised several times, for example [here](): 

```
./../../include/fst/extensions/pdt/shortest-path.h:451:33: error: no member named 'emplace' in 'std::unordered_multimap<fst::internal::ParenState<fst::ArcTpl<fst::TropicalWeightTpl<float> > >, fst::ArcTpl<fst::TropicalWeightTpl<float> >, fst::internal::ParenState<fst::ArcTpl<fst::TropicalWeightTpl<float> > >::Hash, std::equal_to<fst::internal::ParenState<fst::ArcTpl<fst::TropicalWeightTpl<float> > > >, std::allocator<std::pair<const fst::internal::ParenState<fst::ArcTpl<fst::TropicalWeightTpl<float> > >, fst::ArcTpl<fst::TropicalWeightTpl<float> > > > >'
```

[According to cppreference](http://en.cppreference.com/w/cpp/container/unordered_multimap/emplace) `emplace` is a member as of C++11.  Is it possible that this is as "easy" as adding a `-std=c++11` flag for `clang`?  What would i need to change, here?

springmeyer:
> error: no member named 'emplace' in std::unordered_multimap ...
This indicates that a c++11 feature is being used: you can tell since `std::unordered_multimap::emplace` says "since c++11" at http://en.cppreference.com/w/cpp/container/unordered_multimap/emplace. So, the problem is that the compiler has enabled the language feature but the implementation cannot be found. The implementation is in the standard library. So what is happening is that the standard library needs to be upgraded. We should avoid the problem happening in the future by moving our linux builds to use a base docker image (#590). In the meantime what you need to do is to make the `.travis.yml` at `scripts/openfst/1.6.7/.travis.yml` have a matrix item for linux like:

```yml
    - os: linux
      sudo: false
      addons:
        apt:
          sources:
           - ubuntu-toolchain-r-test
          packages:
           - libstdc++-4.8-dev
```

boblannon:
Just wanted to update on why this hasn't been closed/merged: I haven't been able to compile it in a way that a dependent python project, [Pynini](http://pynini.opengrm.org/) expects. I'm trying to build pynini with a fork of the supported code, which [adds the `mason_packages` include and library paths](https://github.com/kylebgorman/Pynini/pull/1/files#diff-2eeaed663bd0d25b7e608891384b7298R37)

It seems like some shared objects are not getting compiled. As a result, some things in pynini work, but some don't. Many of pynini's tests fail, and running some commands causes unexplained kernel crashes.  An example: the serialization to disk (and reading from disk) seem to be broken.  when running `PyniniIOTest` I get:

```
pywrapfst.pyx:2738: FstIOError
--------------------------------------------------------- Captured stderr call ---------------------------------------------------------
ERROR: GenericRegister::GetEntry: dlopen(vector-fst.so, 1): image not found
ERROR: MutableFst::Read: Unknown FST type "vector" (arc type = "standard"): <pywrapfst>
```

I haven't been able to determine how `vector-fst.so` would get produced, but suspect that one or more of the flags for `configure` is preventing that from happening.

boblannon:
closing, because we ultimately decided to go with rust instead of C++ for https://github.com/mapbox/fuzzy-phrase

PR #582, opened 2018-03-17
crosstool-ng 1.23.0

brunoabinader:


springmeyer:
@brunoabinader - nice package. Ideally you’d link the ncurses from mason. This will allow the binary to work on RHEL and Amazon linux.

brunoabinader:
@springmeyer good point :+1: though the ncurses-based view is only used when generating new configurations. Most of the times we're only using the `ct-ng` CLI to  run the build scripts from https://github.com/mapbox/crosstool-configs.

PR #586, opened 2018-03-21
Llvm 6.0.0 release override

### LINKED ISSUE #563: Bundle pyyaml in upcoming LLVM packages
The latest `run-clang-tidy.py` inside LLVM depends on `import yaml` after https://github.com/llvm-mirror/clang-tools-extra/commit/1debd0a44db2a304ee4827aeaf7b977485af6ced.

But `yaml` (aka `pyyaml`) is not part of a default python install and must be installed separately.  This however, breaks the promise of mason packages: which is that all deps should "just work" after being installed and should not require any third party installations.

So, I'm planning to bundle `pyyaml` inside the llvm package to avoid all the downstream users (including > hundreds of CI usages) needing to independently install `pyyaml`

springmeyer:
Prep to create an official LLVM 6.0.0 release (previous 6.0.0 package was a dev version).

Should fix #563 before release.

### DIFF
--- a/scripts/clang++/6.0.0/script.sh
+++ b/scripts/clang++/6.0.0/script.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# For context on this file see https://github.com/mapbox/mason/blob/master/scripts/llvm/base/README.md
+
 # dynamically determine the path to this package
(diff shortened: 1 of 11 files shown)


allieoop:
Issue #563 is fixed and LLVM version 6.0.0 is released.

PR #596, opened 2018-04-16
vtzero 1.0.1

allieoop:
Adding vtzero-1.0.1 build and will `mason trigger` once tests pass.

cc @mapbox/core-tech 
cc @joto


allieoop:
Job triggered and headers are now on s3.

PR #599, opened 2018-04-22
Boost 1.67.0

springmeyer:
Created with:

```bash
git checkout -b boost-1.67.0
./utils/new_boost.sh create 1.67.0 1.66.0
git add scripts/boost*/1.67.0/
git commit -a -m "boost 1.67.0"
./utils/new_boost.sh trigger 1.67.0
```

Then I noticed the boost_python builds failed. Debugging locally I found that the library name changed. Fixed this in bb564ee and then manually triggered a build for just boost_python in: https://travis-ci.org/mapbox/mason/builds/369861321 with ` ./mason trigger boost_libpython 1.67.0`

### DIFF
--- a/scripts/boost/1.67.0/.travis.yml
+++ b/scripts/boost/1.67.0/.travis.yml
@@ -1,0 +1,10 @@
+language: generic
+
+matrix:
+  include:
+    - os: linux
+      sudo: false
+
+script:
+- ./mason build ${MASON_NAME} ${MASON_VERSION}
+- ./mason publish ${MASON_NAME} ${MASON_VERSION}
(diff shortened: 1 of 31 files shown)


springmeyer:
@allieoop - assigned you as a reviewer here. Let's go over this together to check my work / confirm the fix works before merging.

allieoop:
My PR walk-through with @springmeyer helped make sense of the changes he made. The first commit shows the fix he made and that's what we spent time discussing. The second commit is the automated/boilerplate code. 

After I run through a manual test of boost_libpython like we discussed, I think we should be good to merge this!

PR #605, opened 2018-04-30
gzip-hpp 1.0.0

GretaCB:
Per https://github.com/mapbox/gzip-hpp/issues/12, this PR is officially releasing gzip-hpp 1.0.0

cc @allieoop @springmeyer


springmeyer:
Noticed one issue here. The way the packaging is done it is missing one include file: the `include/gzip.hpp` header. Context is that gzip-hpp includes both very specific purpose headers in `include/gzip/*` (to adhere to https://github.com/mapbox/cpp/blob/master/glossary.md#zero-cost-design) and a single header than includes all the others at `include/gzip.hpp`. I'm pretty sure that the `include/gzip.hpp` is getting left out here. This is likely unimportant in the sense that most users downstream should be using the "zero-cost headers" inside `include/gzip/*`. But it could create a  confusing experience for someone using mason that follows the instructions at https://github.com/mapbox/gzip-hpp#usage which recommend doing `#include <gzip.hpp>`. If `include/gzip.hpp` is not present that will be broken and only `#include <gzip/compress.hpp` will work..

allieoop:
Sounds like we can add this tomorrow, but now I'm curious if we even should given that it's easy to point out the zero-cost design benefits of just include the header file with the code that you need. Maybe it would make more sense to update the docs?

springmeyer:
> Maybe it would make more sense to update the docs?

I had this same thought in the back of my mind. Having 2 ways to do something is more error prone. If we went with the "update gzip-hpp" approach however it would be a bit more than updating the docs as the unit tests are using `#include <gzip.hpp>` style.

GretaCB:
Per chat with @allieoop , we'll plan to remove `include/gzip.hpp` since it adds an extra file to manage and some of the files included may not even be needed. Per this discussion, we're also planning on adding a [`npm testpack`](https://github.com/mapbox/node-cpp-skel/blob/master/Makefile#L81-L84) equivalent so we can gutcheck what will be packaged before running `trigger`.

PR #606, opened 2018-04-30
Update publish docs

GretaCB:
Perhaps add specific steps, for example...

The general flow will be:

 - Tag here: `git tag v1.0.0 -a -m "Tagging gzip-hpp v1.0.0"`
 - Push the tag: `git push --tags`
 - Then head over to mason to:
   - Create a new branch
   - Create a `gzip` package
   - Push the branch
   - Test a build with `./mason trigger gzip 1.0.0`

---------
Other additions:
- "Dont for get to merge your PR!"
- Simplify/clarify steps, separate from in-depth context
- Naming conventions
- Add pre-trigger package gutcheck to the doc workflow (similar to [`npm testpack`](https://github.com/mapbox/node-cpp-skel/blob/master/Makefile#L81-L84))
- Note to [update the mason version/gitsha](https://github.com/mapbox/hpp-skel/blob/master/scripts/setup.sh#L6) after adding new package to mason (if using the skels' build setup)

cc @allieoop

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,3 +488,5 @@ We have to override the `mason_cflags` and `mason_ldflags` commands since the re
 ## Troubleshooting
 
 Downloaded source tarballs are cached in `mason_packages/.cache`. If for some reason the initial download failed, but it still left a file in that directory, make sure you delete the partial download there.
+
+## Conventions


allieoop:
@GretaCB - After updating the Mason README I was thinking how the steps you documented above would go well in a *Publishing workflow* section that provides an example workflow which our team uses to tag, build, release, and use new Mason packages. This would be a more appropriate section to have steps liek `git push --tags` and "merge pr" etc. 

I think [this PR](https://github.com/mapbox/mason/pull/608) helps with one of our goals listed here:

* Simplify/clarify steps, separate from in-depth context.

I added a *Prerequisites* section and simplified the *Creating a package* section by separating *Getting started* steps from "this is what a script looks like." I think it's easier to read now but I'll see what you and @springmeyer think after reviewing it.

Also I decided to keep naming conventions out of the docs afterall. I think how someone names a package might belong in a *workflow* section where we document how _we_ like to do things on our team, but not in the core packaging steps. I think leaving this up to each developer and their team is probably ideal.

Adding a pre-trigger package gutcheck to the doc workflow (similar to npm testpack) should be a separate issue. I created that [here](https://github.com/mapbox/mason/issues/609).

I still need to create separate PRs for:
 - [ ] Add a *Publishing workflow* section that provides an example workflow which our team uses to tag, build, release, and use new Mason packages.
 - [ ] Note to update the mason version/gitsha after adding new package to mason (if using the skels' build setup)

GretaCB:
> Also I decided to keep naming conventions out of the docs afterall...I think leaving this up to each developer and their team is probably ideal.

👍 

> Adding a pre-trigger package gutcheck to the doc workflow (similar to npm testpack) should be a separate issue. I created that here.

🙌 

> I still need to create separate PRs for:

Awesome. Once those are open, we can close this PR.

allieoop:
I updated https://github.com/mapbox/mason/pull/608 to include all the changes we discussed. I think we can close this PR now. Since we are leaving out the `# Conventions` header I'm going to close without merging.

PR #607, opened 2018-05-01
Gzip 0.1.0

GretaCB:
Per https://github.com/mapbox/gzip-hpp/pull/26

cc @allieoop @springmeyer


allieoop:
lgtm - the expected header files are there when we build

PR #608, opened 2018-05-04
Mason readme refactor

allieoop:
Giving this README some love with some fresh 👁 👁 👁

My goals in modifying this README are:

* To make it easier to navigate between sections
  - By using a Table of Contents and frequent use of header-links 
  - By reordering sections
* To reduce redundancy
  - By combining instructions for creating header-only and compiled-library packages 
* To reduce ambiguity 
  - By explaining how instructions are just for getting started and that, depending on the complexity of the code being packaged, you may need to do more work
  - Also by listing out questions you should ask yourself before creating a package
* To add useful new sections
  - Releasing a package
  - Using a package
  - Mason internals

Reference: https://github.com/mapbox/mason/pull/606

cc @springmeyer @GretaCB

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,6 @@ Mason works on both **OS X** and **Linux**.
 # Table of Contents
 
 - [Installation](#installation)
-    + [Curl install](#curl-install)
-    + [Submodule](#submodule)
-    + [mason.cmake](#masoncmake)
 - [Configuration](#configuration)
 - [Usage](#usage)
 - [Creating a package](#creating-a-package)
@@ -42,11 +39,11 @@ Mason works on both **OS X** and **Linux**.
 
 There are three recommended ways to install mason, via:
 
-    * curl
-    * a submodule
-    * bundling `mason.cmake`
+* [Curl](#curl)
+* [Submodule](#submodule)
+* [mason.cmake](#masoncmake)
 
-#### Curl install
+#### Curl
 
 To install mason locally:
 
@@ -329,8 +326,8 @@ These are just basic steps to help get you started. Depending on the complexity 
         - url (first parameter): set to the location of your source code archive, e.g. `https://github.com/mapbox/your-lib/archive/v${MASON_VERSION}.tar.gz`
         - checksum (second parameter): set to the checksum you get back after running a checksum function on the source code you want to download. The easiest way to get this checksum is to run `./mason build your-lib 0.1.0` (after setting the `mason_download` url) which will fail with an error message that will contain the correct checksum
     - `mason_compile`
-        - for header-only see []() for an example
-        - for code that needs to be compiled see []() for an example
+        - for header-only see [geometry 0.9.2](https://github.com/mapbox/mason/blob/master/scripts/geometry/0.9.2/script.sh#L19) for an example
+        - for code that needs to be compiled see [zlib 1.2.8](https://github.com/mapbox/mason/blob/master/scripts/zlib/1.2.8/script.sh#L20) for an example
 
     You **may** also need to override the follow Mason functions:
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Mason
 
-Mason is a package manager for C/C++ applications that allows developers to easily install different versions of packages and to streamline build and publishing workflows.
+Mason is a cross-platform, command-line package manager for C/C++ applications.
 
 Mason is like:
 
@@ -26,12 +26,15 @@ Mason works on both **OS X** and **Linux**.
 - [Configuration](#configuration)
 - [Usage](#usage)
 - [Creating a package](#creating-a-package)
-    * [Prerequisites](#prerequisites)
-    * [Getting started](#getting-started)
-    * [System packages](#system-packages)
-- [Script structure](#script-structure)
-- [Mason variables](#mason-variables)
-- [Mason functions](#mason-functions)
+    - [Prerequisites](#prerequisites)
+    - [Getting started](#getting-started)
+    - [System packages](#system-packages)
+- [Releasing a package](#releasing-a-package)
+- [Using a package](#using-a-package)
+- [Mason internals](#mason-internals)
+    - [Mason scripts](#mason-scripts)
+    - [Mason variables](#mason-variables)
+    - [Mason functions](#mason-functions)
 - [Troubleshooting](#troubleshooting)
 
 ## Installation
@@ -63,7 +66,7 @@ Then you can use the `mason` command like: `/tmp/mason install <package> <versio
 
 #### Submodule
 
-Mason can also be added a submodule to your repository. This is helpful for other contributors to get set up quickly.
+Mason can also be added as a submodule to your repository. This is helpful for other contributors to get set up quickly.
 
 Optionally a convention when using submodules, is to place the submodule at a path starting with `.` to make the directory hidden to most file browsers. If you want your mason folder hidden then make sure to include the final part of the following command `.mason/` so your submodule path has the leading `.` instead of just being `mason/`.
 
@@ -107,7 +110,7 @@ Then in your `CmakeLists.txt` install packages like:
 mason_use(<package name> VERSION <package version> HEADER_ONLY)
 ```
 
-Note: Leave out `HEADER_ONLY` if the package is a [pre-compiled library](https://github.com/mapbox/cpp/blob/master/glossary.md#precompiled-library). You can see if a package is `HEADER_ONLY` by looking inside the `script.sh` for `MASON_HEADER_ONLY=true` like https://github.com/mapbox/mason/blob/68871660b74023234fa96d482898c820a55bd4bf/scripts/geometry/0.9.0/script.sh#L5
+_Note: Leave out `HEADER_ONLY` if the package is a [pre-compiled library](https://github.com/mapbox/cpp/blob/master/glossary.md#precompiled-library). You can see if a package is `HEADER_ONLY` by looking inside the `script.sh` for `MASON_HEADER_ONLY=true` like https://github.com/mapbox/mason/blob/68871660b74023234fa96d482898c820a55bd4bf/scripts/geometry/0.9.0/script.sh#L5_
 
 ## Configuration
 
@@ -138,7 +141,7 @@ The `command` can be one of the following
 * [link](#link) - creates symlinks for packages in `mason_packages/.link`
 * [trigger](#trigger) - trigger a build and publish operation on Travis CI
 
-### install
+#### install
 
 ```bash
 $ mason install libuv 0.11.29
@@ -155,8 +158,7 @@ Next, Mason checks whether there are pre-built binaries available in the S3 buck
 
 If no pre-built binaries are available, Mason is going to build the library according to the script in the `mason_packages/.build` folder, and install into the platform- and library-specific directory.
 
-
-### remove
+#### remove
 
 ```bash
 $ mason remove libuv 0.11.29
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,38 +1,50 @@
 # Mason
 
-Mason can compile C/C++ code, publish packages to S3, and install those packages via binaries.
+Mason is a package manager for C/C++ applications that allows developers to easily install different versions of packages and to streamline build-and-publish workflows.
 
-Mason does all its work locally inside `./mason_packages` and therefore does not require privileges for, or conflict with, system-wide packages.
+Mason is like:
```
GretaCB: This like/unlike is super helpful

README.md
```diff
@@ -1,38 +1,50 @@
...
 
-Mason is unlike all of the above package managers because:
+[![Build Status](https://travis-ci.org/mapbox/mason.svg?branch=master)](https://travis-ci.org/mapbox/mason)
 
-  - Mason runs on both Linux and OS X and creates a single set of binaries that work on >= OS X 10.8 and >= Ubuntu Precise (rather than building binaries per version).
-  - Mason strongly prefers static libraries over shared libraries
-  - Mason has zero understanding of dependency trees: it leaves complete control to the developer for how packages relate.
-  - Mason does not depend on any specific runtime language (like python, node.js or ruby). It is a just a few bash scripts.
-  - Mason depends on [travis.ci](https://travis-ci.org) for creating and publishing binaries.
+# Table of Contents
```
GretaCB: ⭐️ noice

README.md
```diff
@@ -1,38 +1,50 @@
...
+    * [System packages](#system-packages)
+- [Troubleshooting](#troubleshooting)
+- [Conventions](#conventions)
 
 ## Installation
 
-There are three recommended ways to install mason: via curl, via a submodule, or via bundling `mason.cmake`
+There are three recommended ways to install mason, via:
+
+    * curl
```
GretaCB: Looks like adding tabs automatically converts to a code block format
```
like this
```

Not sure if that was intended. Bullet points in markdown automatically indent.

GretaCB: p.s. Love how this intros and sets the stage for the three ways to install

README.md
```diff
@@ -231,172 +259,159 @@ curl -s -i https://api.travis-ci.org/auth/github \
 
 For details see https://docs.travis-ci.com/user/triggering-builds and https://github.com/travis-ci/travis.rb#readme
 
-## Writing build scripts
+## Creating a package
 
-Each version of each package has its own directory in `scripts/${package}/${version}`, e.g. [`scripts/libpng/1.6.28`](https://github.com/mapbox/mason/tree/master/scripts/libpng/1.6.28).
+### Prerequisites
```
GretaCB: Gorg

README.md
```diff
@@ -231,172 +259,159 @@ curl -s -i https://api.travis-ci.org/auth/github \
...
-If you are creating a package for previously unpackaged software start by creating a new directory for your `${package}/${version}` from within your mason checkout.
+**Are there previous versions already published?**
 
-For example if you want to name your package `yourlib` and it is version `1.0.0` you would do:
+Check the [list of packages](https://github.com/mapbox/mason/tree/master/scripts) to see if a previous version of your package exists. It's helpful to copy scripts from a previous version rather than creating code from scratch.
 
-```bash
-mkdir -p scripts/yourlib/1.0.0
-```
+If there is no previous version of your package, it is recommended to copy a simple package that has mostly boiler plate code:
```
GretaCB: Having examples of both `version tag` and `gitsha` is super helpful here.

README.md
```diff
@@ -231,172 +259,159 @@ curl -s -i https://api.travis-ci.org/auth/github \
...
-What you put in those files depend on what type of package you are creating. See:
+Every package needs to tell Mason where to download the code that it will build and publish, for example:
 
- - [Header-only package](#header-only-package)
- - [Pre-compiled library package](#pre-compiled-library-package)
+ - `https://github.com/mapbox/geometry.hpp/archive/v0.9.2.tar.gz` for a Github release: [geometry 0.9.2](https://github.com/mapbox/geometry.hpp/releases/tag/v0.9.2)
+ - `https://github.com/mapbox/geometry.hpp/archive/b0e41cc5635ff8d50e7e1edb73cadf1d2a7ddc83.zip` for pre-release code hosted on Github: [geometry b0e41cc](https://github.com/mapbox/geometry.hpp/tree/b0e41cc5635ff8d50e7e1edb73cadf1d2a7ddc83)
 
-### Creating new versions of packages
+Note: Your code doesn't need to be hosted on Github in order for Mason to work. Your code can be hosted anywhere. Another common location is [SourceForge](#https://sourceforge.net/).
```
GretaCB: Is there a link we can add here to point to where this can be configured?

README.md
```diff
@@ -231,172 +259,159 @@ curl -s -i https://api.travis-ci.org/auth/github \
...
-#### Header-only package
+2. Add scripts for building and publishing your package.
 
-For a header-only library, a good example to copy from is the `geometry` package. You can copying the `geometry` package:
+    Each package must have the following two files: `script.sh` and `.travis.yml`. Copy these two files from a previous version of your package, or, if one doesn't exist, from a package of the same type (either compiled-code or header-only) that contains mostly boiler-plate code. For more details on where to look, see [Prerequisites](#prerequisites).
 
-```bash
-cp -r scripts/geometry/0.9.1 scripts/yourlib/1.0.0
-```
+    For example, if you're creating the first package of a library named `your-lib` that is version `0.1.0` you would copy scripts from the `geometry` package:
```
GretaCB: > of a header-only library
to highlight why copying geometry lib?

README.md
```diff
@@ -231,172 +259,159 @@ curl -s -i https://api.travis-ci.org/auth/github \
...
+    You **must** set the follow Mason variables:
 
-Then you will also need to update:
+    - `MASON_NAME`: set to the name of your package, e.g. `your-lib`
+    - `MASON_VERSION`: set to the package version, e.g. `0.1.0`
+    - `MASON_BUILD_PATH`: set to the location Mason will use to store header files or binaries before it packages and publishes them
 
- - The github url on line `11` to match your download url
- - The hash on line `12`. To do this, it is easiest to run `./mason build yourlib version` and it will fail, but the error message will contain the correct hash. Copy the hash and put it on line `12`.
+    You **may** also need to set the follow Mason variables:
```
GretaCB: Dig the RFC use :)

README.md
```diff
@@ -231,172 +259,159 @@ curl -s -i https://api.travis-ci.org/auth/github \
...
 
-Now make any adjustments needed to the `mason_compile` section, modifying the copying of headers.
+    You **must** override the follow Mason functions:
 
-#### Pre-compiled library package
+    - `mason_load_source`: you must call `mason_download` and update its parameters:
+        - url (first parameter): set to the location of your source code archive, e.g. `https://github.com/mapbox/your-lib/archive/v${MASON_VERSION}.tar.gz`
+        - checksum (second parameter): set to the checksum you get back after running a checksum function on the source code you want to download. The easiest way to get this checksum is to run `./mason build your-lib 0.1.0` (after setting the `mason_download` url) which will fail with an error message that will contain the correct checksum
+    - `mason_compile`
+        - for header-only see []() for an example
```
GretaCB: link link

README.md
```diff
@@ -231,172 +259,159 @@ curl -s -i https://api.travis-ci.org/auth/github \
...
+    # compressed, you can also use `mason_extract_tar_bz2` instead.
     mason_extract_tar_gz
 
-    export MASON_BUILD_PATH=${MASON_ROOT}/.build/libuv-${MASON_VERSION}
+    # This variable contains the path to the unpacked folder inside the `.build` directory.
+    export MASON_BUILD_PATH=${MASON_ROOT}/.build/${MASON_NAME}-${MASON_VERSION}
 }
-```
-
-In that function, you should use `mason_download` as a shortcut to download the tarball. The second argument to is a hash generated with `git hash-object` and used to verify that the source code downloaded matches the expected file. The function also caches downloaded tarballs in the `mason_packages/.cache` folder. You can get the hash object by running `./mason build <package> <version>` locally and this should fail with a message of the expected hash.
```
GretaCB: Worth it to retain
> The function also caches downloaded tarballs in the `mason_packages/.cache` folder.

here still? The only other place cache is mentioned is at the bottom of the doc.

README.md
allieoop: Not intended. Fixed!

README.md
allieoop: Added links. Thnx.

README.md
allieoop: @GretaCB - I removed it from here because it wasn't helping my understanding of how we git the checksum back and I wanted to reduce clutter in the **Getting started** section. I could add it to the cache section at the bottom and list everything we store there. What do you think? Did it actually help your understanding of the process to create a package?

GretaCB: > Did it actually help your understanding of the process to create a package?

Great question. My assumption is that this [`Script structure`](https://github.com/mapbox/mason/blob/0800a52406e28955c3ee3204a35ca3c9a64302d6/README.md#script-structure) section is the main place to go for details on what each function does. And part of what `mason_load_source` does is add to `/.cache`. I can see it being handy to have that in one place for reference.

What are your thoughts on `Script Struture` being a standalone section, but _referenced_ for Getting started?

README.md
allieoop: Yeah, I'd like to figure out how we do this and where an example is. I'll try to find one.

README.md
allieoop: I agree. I'll add it to the `Script structure` section.

I think moving `Script structure` out of the `Creating a package` section is good idea too. Also: `Mason variables` and `Mason functions` can be moved up a level as well. 👍

README.md
```diff
@@ -231,172 +257,271 @@ curl -s -i https://api.travis-ci.org/auth/github \
...
 
-What you put in those files depend on what type of package you are creating. See:
+Note: Your code doesn't need to be hosted on Github in order for Mason to work. Your code can be hosted anywhere. Another common location is [SourceForge](#https://sourceforge.net/).
 
- - [Header-only package](#header-only-package)
- - [Pre-compiled library package](#pre-compiled-library-package)
+### Getting started
 
-### Creating new versions of packages
+These are just basic steps to help get you started. Depending on the complexity of building your code, you might have to review the [Script structure](#script-structure) section to get a better idea of how to further configure Mason to be able to create your package.
```
mapsam: @allieoop this `script structure` link doesn't seem to be working for me - perhaps the section title changed?

README.md
```diff
@@ -231,172 +257,271 @@ curl -s -i https://api.travis-ci.org/auth/github \
...
+
+3. Create a package.
+
+    We recommend working in a new branch before creating a pacakge.
+
+    `git checkout -b my-version-0.1.0`
+
+    Follow the instructions in the [Getting started](#getting-started) section for creating a new package.
+
+4. Build your package.
```
mapsam: Worth noting that the `./mason build <package> <version>` will output a checksum error? I find myself searching for these docs pretty regularly.

allieoop: @mapsam - If the *Getting started* instructions are followed in step 3 then they should not run into that error. It's also mentioned there how we use that error to get the correct checksum. Looking forward to Mason v2 where I imagine we won't have to do this step. :)

README.md
allieoop: good find. will update.

README.md
allieoop: If step 3 is followed, in particular the **Getting started** instructions that step 3 links to, this is unnecessary.

mapsam: 👍 didn't see it linked to the "getting started" section - sounds good!

README.md
```diff
@@ -231,172 +257,278 @@ curl -s -i https://api.travis-ci.org/auth/github \
...
-#### Script structure
+    The location of your package depends on how you configured Mason (see [Configuration](#configuration)) and the type of package you published. For example if you published your package to the default S3 bucket (`mason-binaries`) and it's a header-only package, you can find your package here:
 
-This `script.sh` is structured like:
+    `aws s3 ls s3://mason-binaries/headers/my_new_package/0.1.0.tar.gz`
+
+## Using a package
+
+1. First make sure you followed the [Mason installation](#installation) instructions.
+
```
millzpaugh: @allieoop I would say something like: 

Mason has two clients for installing packages - 
1) The Mason cli tool - you can follow the install instructions in the Mason docs <link> The cli install client is part of the Mason project itself. 
2) Mason-js - you can follow its docs here: https://github.com/mapbox/mason-js. Mason-js is a separate node JS client. 

There are examples of projects using the cli tool and mason-js. 
* [hpp-skell](https://github.com/mapbox/hpp-skel) uses the cli tool. This install client requires you to update your Mason version in `scripts/setup.sh`
 * [node-cpp-skel](https://github.com/mapbox/node-cpp-skel) uses mason-js, a JS client that pulls packages directly from S3.

The install command syntax will differ depending on the client you use.

PR #611, opened 2018-05-17
Public Valhalla 2.4.9

danpat:
# DO NOT MERGE YET

Packaging for Valhalla 2.4.9 binaries and libraries.  Currently only includes runtime dependencies, does not build/bundle tile-creation, and service binaries.

## TODO:

- [x] package lz4 with mason
- [x] ~~package prime_server with mason~~ (will not include services in this package version)
- [x] use libcurl from mason
- [x] Figure out how to link in openssl

## Notes

To build/test this locally, clone this branch, then do `./mason build valhalla 2.4.9`

PR #614, opened 2018-05-22
Add libspatialite 4.3.0a

oxidase:
Required for https://github.com/mapbox/mason/pull/611


oxidase:
Not needed as `libvalhalla` is configured without data tools.

PR #619, opened 2018-06-22
Add geojsonvt v6.4.0

mourner:


springmeyer:
Whoops, just noticed a glitch in the `.travis.yml` here, that is probably copy/paste of a previous bug. @mourner change the .travis to be like: https://github.com/mapbox/mason/blob/master/scripts/geometry/0.9.2/.travis.yml

That moves the `sudo` into the matrix and will ensure that there is only one job that runs on travis for the `./mason trigger ...`

mourner:
@springmeyer I already ran `mason trigger`... Should I re-run after the fix, or just add it for the next copy-paste?

springmeyer:
> @springmeyer I already ran mason trigger... Should I re-run after the fix, or just add it for the next copy-paste?

Pretty sure that s3 put are atomic - so your first trigger should be fine. But good to have this fixed to avoid confusion in the future. Merge away!

PR #620, opened 2018-06-22
[WIP] Fix preservation of symlinks in llvm subpackages

springmeyer:
I noticed that the `c++` dir symlink on OS X in the clang++ and clang-tidy packages is not behaving predictably across packaged versions. I think it must be different behavior between `cp -r` in the latest coreutils and `cp -r`  in the default bsd tools on OS X. Need to dig deeper. Putting this up as a reminder of the problem that needs fixing (keeping the symlink)

Currently, in the LLVM 6.0.0 packages clang-tidy's c++ directory is a symlink while clang++ is not (they both should be).


springmeyer:
Going to fix this via no longer symlinking this `c++` directory: #650

PR #636, opened 2018-07-22
Added Benchmark 1.4.1

brunoabinader:
@springmeyer had to update clang++ to include `llvm-nm`, needed by benchmark.

### DIFF
--- a/scripts/benchmark/1.4.1-cxx11abi/.travis.yml
+++ b/scripts/benchmark/1.4.1-cxx11abi/.travis.yml
@@ -1,0 +1,28 @@
+language: cpp
+
+sudo: false
+
+compiler: clang
+
+addons:
+  apt:
+    update: true
+    sources:  [ 'george-edison55-precise-backports' ]
+    packages: [ 'cmake', 'cmake-data' ]
+
+matrix:
+  exclude:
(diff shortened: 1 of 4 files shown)


springmeyer:
@brunoabinader - can you document more details on the error you hit exactly? And did you publish a new clang++ package to fix this? If so what version?

brunoabinader:
@brunoabinader - can you document more details on the error you hit exactly? And did you publish a new clang++ package to fix this? If so what version?

Sure - benchmark 1.4.1 [requires llvm-nm](https://github.com/google/benchmark/blob/c45f01866bdfe7c62a83b51dfb974acb7fa5cab6/cmake/llvm-toolchain.cmake#L4-L5), which wasn't available on the mason package for clang++ 5.0.0.

I did not need to recompile LLVM for that - Our LLVM package (used to generate clang++) already provides `llvm-nm` binary, so I just needed to [add it for generating clang++ too](https://github.com/mapbox/mason/pull/636/files#diff-7d9048c66bbef4edf48d395aa46fd651R17). The clang++ package in question was 5.0.0, as it is the [mason default](https://github.com/mapbox/mason/blob/master/utils/toolchain.sh).

PR #637, opened 2018-07-24
Add vector tile v1.0.2

nagineni:
SSIA


nagineni:
Already ran 'trigger', and it was successful.

Issue #230, opened 2016-08-16
Symbolic link command should use `-f` flag consistently 


artemp:
https://travis-ci.org/mapnik/mapnik/jobs/152731996#L486

```
ln: failed to create symbolic link `/home/travis/build/mapnik/mapnik/mason_packages/.link/mason.ini': File exists
```

I'm getting bunch of errors like above which are temperamental. I think the solution is to always unlink before creating a symbolic link via `-f` flag. 

/cc @springmeyer @kkaefer

springmeyer:
The odd thing about this bug is that `mason_packages/.link/mason.ini` is not a file that should every get created as far as I can tell. It may be that `mason_config` is getting called while the current working directory is not correct (or there is a race condition). The `mason_config` functionality was added in https://github.com/mapbox/mason/commit/bfc612dbaaec13710a202268a0ecb7f494a6671f but the bug is likely from https://github.com/mapbox/mason/pull/216 (which fixed other linking problems).

kkaefer:
@springmeyer the mason.ini files are created as part of https://github.com/mapbox/mason/pull/176

springmeyer:
I see the problem now. The `mason.ini` for each package is being linked to the same place, so each packages tries to create `.mason_packages/.link/mason.ini`. This is because of the usage of `lndir` which creates "a shadow directory of symbolic links to another directory tree". The `lndir` command is used since it is way faster than `ln -s` per file. But with `lndir` we don't have control over ignoring files that would clobber each other across packages. Hmmmm.

springmeyer:
@kkaefer what do you think about storing `mason.ini` at `${MASON_PREFIX}/<package-name>.ini` instead of `${MASON_PREFIX}/mason.ini`. That would be a breaking change of course, but it would allow the ini config to be used within a linked/collapsed directory structure.

kkaefer:
While we could rename it, we decided to [deprecate linking](https://github.com/mapbox/mason/issues/377)

lightmare:
This can be closed as per the above-mentioned PR.

springmeyer:
Thanks @lightmare!

PR #641, opened 2018-08-02
mason link: overwrite existing

lightmare:
Refs #230, mapnik/mapnik#3955

Not rewriting symlinks leaves links to files from older/different package version in place.

First I rewrote the bash fallback, it was very slow. When symlinking boost headers, for me it ran ~35s when creating the directory tree, and ~3s when only checking existence, not creating/updating any links. The new function runs ~4s when creating or updating the directory tree.

Then I installed `lndir` only to learn it doesn't have any option to the effect of `--force`. So I deemed it unfit for the task. GNU `cp` has option `--symbolic-link`, which runs a bit slower than `lndir`, but still a lot faster than the new bash function. So that is used if available.

**changes in behaviour**
- existing symlinks/files are overwritten
- files in the top-level directory are not symlinked (`mason.ini`, etc.)

It should be noted that I used `find` options `-mindepth` and `-maxdepth`. They're not specified by POSIX, and although OSX does recognize them, I'm not entirely sure they're ok to use here.

### DIFF
--- a/mason.sh
+++ b/mason.sh
@@ -368,24 +368,34 @@ function mason_clean {
 }
 
 function bash_lndir() {
-    oldifs=$IFS
-    IFS='
-    '
-    src=$(cd "$1" ; pwd)
-    dst=$(cd "$2" ; pwd)
-    find "$src" -type d |
-    while read -r dir; do
-            mkdir -p "$dst${dir#$src}"
-    done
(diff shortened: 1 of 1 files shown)


springmeyer:
👍 Thanks for fixing/improving this @lightmare

PR #645, opened 2018-08-09
Add SQLite 3.24.0 optimized for binary size

jfirebaugh:
Using both `-Os` and `-flto` (which we've [found to substantially reduce binary size](https://github.com/mapbox/mapbox-gl-native/runs/10667579)).

Testing over in https://github.com/mapbox/mapbox-gl-native/pull/12596.

This could serve as an example for https://github.com/mapbox/cpp/issues/22


jfirebaugh:
Gave up on LTO. It probably doesn't make too much of a difference anyway, given that SQLite compiles as a single large .c file.

PR #647, opened 2018-08-15
Change download location for bzip2 -- the previous one no longer works

apendleton:
The bzip.org domain has expired and been replaced by adware, so the install script that tries to download it now fails (it downloads something, but the hash check fails). This PR switches the download location to use the Fedora project's mirror of the source package -- it hashes identically, and has been there since 2010, so I don't think it's going anywhere.


springmeyer:
Thanks @apendleton 

>  (it downloads something, but the hash check fails).

Yay, the hash check actually helped with something for once (rather than just annoying us).  /cc @alliecrevier since we've talked about this and how I've wondered if we should remove the hash check. This is case where it helped.

PR #653, opened 2018-08-24
Add more platforms to boost libdatetime builds.

danpat:


scripts/boost_libdate_time/1.66.0/.travis.yml
```diff
@@ -1,19 +1,42 @@
-language: generic
+sudo: false
```
springmeyer: this `sudo:false` needs to be nested inside each specific matrix it should apply to otherwise it will trigger duplicate builds in travis due to how travis + inheritance works (since this will inherit from https://github.com/mapbox/mason/blob/master/.travis.yml)

springmeyer:
@danpat still plan/need to land this or can it be closed?

PR #630, opened 2018-07-11
update gitsha section in readme

alliecrevier:
Improvement on wording around which hash we're talking about when creating a new mason package for code managed in git.

/cc @millzpaugh

PR #658, opened 2018-09-19
Revert "update gitsha section (#630)"

springmeyer:
This reverts commit 6352cf806a30b4348ee1c3ce1755837de2389d81 and https://github.com/mapbox/mason/pull/630.

It seems (from reports from @julianrex ) that the `git ls-remote` approach does not actually generate a valid sha value. Additionally not all mason packages are hosted on github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,7 +325,7 @@ These are just basic steps to help get you started. Depending on the complexity 
 
     - `mason_load_source`: you must call `mason_download` and update its parameters:
         - url (first parameter): set to the location of your source code archive, e.g. `https://github.com/mapbox/your-lib/archive/v${MASON_VERSION}.tar.gz`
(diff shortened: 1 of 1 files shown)


springmeyer:
@julianrex hit this doc bug. @julianrex could you:

- [ ] confirm this approach works for you?
- [ ] comment if there are any other doc updates you think are needed for knowing how to publish new versions of packages?

julianrex:
After setting up the Travis token, I triggered the build with:
```
MASON_TRAVIS_TOKEN=<token> ./mason trigger supercluster 0.3.1
```
It might be worth suggesting this in the readme?

springmeyer:
@julianrex 

> After setting up the Travis token, I triggered the build with:
> MASON_TRAVIS_TOKEN=<token> ./mason trigger supercluster 0.3.1
> It might be worth suggesting this in the readme?

It is mentioned at https://github.com/mapbox/mason#trigger. Where you looking for it somewhere else that would have been more noticeable?

PR #666, opened 2018-11-06
Icu 63.1 min static data

kevinkreiser:
This is a build of ICU with `-Os` and with the data packaging set to put the data in a static lib. This also removes a significant amount of the data. It narrows down the datasets to just: `locales`, `unit`, `currency` and a couple basic things in `misc`. Additionaly for those datasets the only locales supported are:

    da de en eo es fi fr he id it ko my nl pl pt pt_PT ro ru sv tr uk vi zh zh_Hans

With all of these changes the total data size of the data library is down from 26MB to 3.1MB. Which is quite convenient for cross platform libraries that have to ship the same code to a variety of platforms.

There was a slight issue with the standard download/unpacking workflow. ICU does not ship the "code" used to build the data in the same targz as the code. Because of this the package must download the code and the data-code and splice them together (to get the static library with data built-in). What I've done is updated the downloading function with the ability to add a suffix to the file you downloaded. This allows the unpacking functions to then use the same suffix and thefore avoids destroying the cache of the other files that were downloaded for the package in question. Because of this change I am wondering if I should update the mason version from `0.18.0` to `0.19.0` or if even just new packages in mason also obviate the version update? @springmeyer or maybe @brunoabinader can you advise?

### DIFF
--- a/scripts/icu/63.1-min-static-data/script.sh
+++ b/scripts/icu/63.1-min-static-data/script.sh
@@ -33,7 +33,7 @@ function mason_prepare_compile {
         #  Also instead of using the regular build steps, we could use a dedicated built target that just builds the tools
         mason_substep "Moving host ICU build directory to ${MASON_ROOT}/.build/icu-host"
         rm -rf ${MASON_ROOT}/.build/icu-host
-        cp -R ${MASON_BUILD_PATH}/source ${MASON_ROOT}/.build/icu-host
+        cp -R ${MASON_BUILD_PATH}/icu4c/source ${MASON_ROOT}/.build/icu-host
     fi
 }
 
--- a/scripts/icu/63.1-min-static-data/script.sh
+++ b/scripts/icu/63.1-min-static-data/script.sh
@@ -63,17 +63,17 @@ function trim_data {
         # make a local version and remove the line continuations
         l=$(echo $f | sed -e "s/\(.*\)files.mk/\1local.mk/g")
         cp -rp $f $l
-        sed -i -e :a -e '/\\$/N; s/\\\n//; ta' $l
+        sed -i'' -e :a -e '/\\$/N; s/\\\n//; ta' $l
 
         # if its locale, unit or currency we use our supported lang list
         if [ "$(echo $l | grep -cE 'locales|unit|curr')" -eq 1 ]; then
-            sed -i -e '/^#/!s/SOURCE.*=.*txt/SOURCE = da.txt de.txt en.txt eo.txt es.txt fi.txt fr.txt he.txt id.txt it.txt ko.txt my.txt nl.txt pl.txt pt.txt pt_PT.txt ro.txt ru.txt sv.txt tr.txt uk.txt vi.txt zh.txt zh_Hans.txt/g' $l
+            sed -i'' -e '/^#/!s/SOURCE.*=.*txt/SOURCE = da.txt de.txt en.txt eo.txt es.txt fi.txt fr.txt he.txt id.txt it.txt ko.txt my.txt nl.txt pl.txt pt.txt pt_PT.txt ro.txt ru.txt sv.txt tr.txt uk.txt vi.txt zh.txt zh_Hans.txt/g' $l
         # if its misc we need a couple of things
         elif [ $(echo $l | grep -cF "misc") -eq 1 ]; then
-            sed -i -e '/^#/!s/SOURCE.*=.*txt/SOURCE = numberingSystems.txt icuver.txt icustd.txt/g' $l
+            sed -i'' -e '/^#/!s/SOURCE.*=.*txt/SOURCE = numberingSystems.txt icuver.txt icustd.txt/g' $l
         # otherwise the scuttle the whole thing
         else
-            sed -i -e '/^#/!s/SOURCE.*=.*txt/SOURCE = /g' $l
+            sed -i'' -e '/^#/!s/SOURCE.*=.*txt/SOURCE = /g' $l
         fi
     done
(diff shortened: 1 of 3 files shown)


kevinkreiser:
nice pull request number! :imp:

scripts/icu/63.1-min-static-data/script.sh
```diff
@@ -0,0 +1,161 @@
...
+        source
+
+    mason_extract_tar_gz
+
+    mason_download \
+        http://download.icu-project.org/files/icu4c/63.1/icu4c-63_1-data.zip \
+        a0eb02a525822ea5ab409e04cf280bacccbb5d4d \
+        data
+
+    mason_extract_zip -d ${MASON_NAME}/source
```
kevinkreiser: prior to the `SLUG_SUFFIX` this function would have zapped the cache for the previous download.

scripts/icu/63.1-min-static-data/script.sh
```diff
@@ -0,0 +1,161 @@
...
+        # TODO: Copies a bunch of files to a kind of orphaned place, do we need to do something to clean up after the build?
+        #  Copying the whole build directory is the easiest way to do a cross build, but we could limit this to a small subset of files (icucross.mk, the tools directory, probably a few others...)
+        #  Also instead of using the regular build steps, we could use a dedicated built target that just builds the tools
+        mason_substep "Moving host ICU build directory to ${MASON_ROOT}/.build/icu-host"
+        rm -rf ${MASON_ROOT}/.build/icu-host
+        cp -R ${MASON_BUILD_PATH}/source ${MASON_ROOT}/.build/icu-host
+    fi
+}
+
+function trim_data {
```
kevinkreiser: this function is what "patches" the ICU data files so that we only build some of the datasets and only for some of the locales. i could have used a patch file but ICU docs say to copy files and modify those so I regex'd my way there instead.

scripts/icu/63.1-min-static-data/script.sh
```diff
@@ -0,0 +1,161 @@
...
+    
+    CFLAGS="${CFLAGS:-} ${ICU_CORE_CPP_FLAGS} ${ICU_MODULE_CPP_FLAGS} -fvisibility=hidden $(icu_debug_cpp) -Os"
+    CXXFLAGS="${CXXFLAGS:-} ${ICU_CORE_CPP_FLAGS} ${ICU_MODULE_CPP_FLAGS} -fvisibility=hidden $(icu_debug_cpp) -Os"
+
+    echo "Configuring with ${MASON_HOST_ARG}"
+
+    ./configure ${MASON_HOST_ARG} --prefix=${MASON_PREFIX} \
+    $(icu_debug_configure) \
+    $(cross_build_configure) \
+    --with-data-packaging=static \
```
kevinkreiser: this is the flag that makes it so that the data is included in the static library itself as opposed to a separate file

scripts/icu/63.1-min-static-data/script.sh
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env bash
+
+# Build ICU common package (libicuuc.a) with data file separate and with support for legacy conversion and break iteration turned off in order to minimize size
+
+MASON_NAME=icu
+MASON_VERSION=63.1-min-static-data
+MASON_LIB_FILE=lib/libicuuc.a
```
kevinkreiser: @springmeyer can one put more than one library on this line?

springmeyer: This is used to check that the install worked. It only supports a single file and can be anything that exists after the build completes. So, the way you have it here is 👍. No need (or ability) to point to multiple files and not pointing to the other libraries  is okay - they will still get installed.

scripts/icu/63.1-min-static-data/script.sh
springmeyer: 👌

scripts/icu/63.1-min-static-data/script.sh
```diff
@@ -0,0 +1,160 @@
...
+
+function mason_compile_base {
+    pushd  ${MASON_BUILD_PATH}/source
+    
+    # trim out a bunch of the data so that the data static library is as small as possible
+    trim_data
+
+    # Using uint_least16_t instead of char16_t because Android Clang doesn't recognize char16_t
+    # I'm being shady and telling users of the library to use char16_t, so there's an implicit raw cast
+    ICU_CORE_CPP_FLAGS="-DU_CHARSET_IS_UTF8=1 -DU_CHAR_TYPE=uint_least16_t"
```
springmeyer: @kevinkreiser this `U_CHAR_TYPE` customization (and the below `-DUCHAR_TYPE=char16_t` in the `mason_cflags`) was done by @ChrisLoer for MBGL and may not benefit your usecase, so I would consider removing it unless you specific think it is helpful for your usecase.

ChrisLoer: FWIW, mbgl-core is no longer using mason for ICU, so totally fine for the mason build to stop using the `U_CHAR_TYPE` customization.

kevinkreiser:
@springmeyer youre comments should now be addressed. can i get another look?

scripts/icu/63.1-min-static-data/script.sh
kevinkreiser: indeed, we looked at vendoring as well (which i think we want to do long term) but decided against biting the bullet of figuring out how to vendor the data into the lib (so we can ship without having to bundle the data separately)

PR #673, opened 2018-11-20
add one missing data resource back to icu63.1-min-static-data-1

kevinkreiser:
There was a slight mistake in the last PR for this library. It was missing one key data file that allows the library to properly choose the plural/singlar form for units when doing unit formatting. Other than that this is a copy of the previous package.

### DIFF
--- a/scripts/icu/63.1-min-static-data-1/script.sh
+++ b/scripts/icu/63.1-min-static-data-1/script.sh
@@ -1,0 +1,150 @@
+#!/usr/bin/env bash
+
+# Build ICU common package (libicuuc.a) with data file separate and with support for legacy conversion and break iteration turned off in order to minimize size
+
+MASON_NAME=icu
+MASON_VERSION=63.1-min-static-data-1
+MASON_LIB_FILE=lib/libicuuc.a
+#MASON_PKGCONFIG_FILE=lib/pkgconfig/icu-uc.pc
+
+. ${MASON_DIR}/mason.sh
+
+MASON_BUILD_DEBUG=0 # Enable to build library with debug symbols
+MASON_CROSS_BUILD=0
+
+function mason_load_source {
+    # you can't get the data and the code together in one place except when you clone the repo
+    export MASON_BUILD_PATH=${MASON_ROOT}/.build/${MASON_NAME}-${MASON_VERSION}
+    if [[ ! -d ${MASON_BUILD_PATH} ]]; then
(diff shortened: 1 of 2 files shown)


scripts/icu/63.1-min-static-data-1/script.sh
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env bash
+
+# Build ICU common package (libicuuc.a) with data file separate and with support for legacy conversion and break iteration turned off in order to minimize size
+
+MASON_NAME=icu
+MASON_VERSION=63.1-min-static-data-1
```
kevinkreiser: first difference from the `63.1-min-static-data` package is the name change here, adding a `-1` to the end to denote the change.

scripts/icu/63.1-min-static-data-1/script.sh
```diff
@@ -0,0 +1,150 @@
...
+        # make a local version and remove the line continuations
+        l=$(echo $f | sed -e "s/\(.*\)files.mk/\1local.mk/g")
+        sed -e :a -e '/\\$/N; s/\\\n//; ta' $f > $l
+
+        # if its locale, unit or currency we use our supported lang list
+        if [ "$(echo $l | grep -cE 'locales|unit|curr')" -eq 1 ]; then
+            sed -i'' -e '/^#/!s/SOURCE.*=.*txt/SOURCE = da.txt de.txt en.txt eo.txt es.txt fi.txt fr.txt he.txt id.txt it.txt ko.txt my.txt nl.txt pl.txt pt.txt pt_PT.txt ro.txt ru.txt sv.txt tr.txt uk.txt vi.txt zh.txt zh_Hans.txt/g' $l
+        # if its misc we need a couple of things
+        elif [ $(echo $l | grep -cF "misc") -eq 1 ]; then
+            sed -i'' -e '/^#/!s/SOURCE.*=.*txt/SOURCE = plurals.txt numberingSystems.txt icuver.txt icustd.txt pluralRanges.txt/g' $l
```
kevinkreiser: this is the only other difference in this package. we've added `plurals.txt` to the list of data to keep in the library. it turns out this is required to get the library to spit out `1 mile` instead of `1 miles`. :man_facepalming:

scripts/icu/63.1-min-static-data-1/script.sh
springmeyer: 👍

scripts/icu/63.1-min-static-data-1/script.sh
springmeyer: wow, nice work figuring this out 🙌

PR #678, opened 2019-02-13
Postgres 11 w/ Readline

ingalls:
Introduce support for Postgres 11 using mason managed readline library.


ingalls:
O_o @springmeyer huge apology, I actually missed the delivery of the `libedit` version. That should work just fine. I'm going to close out this PR and look at updating to `10.3` with the `libedit` support and move from there to potentially cut an `11` version with `libedit`

PR #689, opened 2019-06-28
Postgres 11.4 Sketch

ingalls:
Use the postgres 10.3 install script as a base for installing postgres 11.4

cc/ @springmeyer


springmeyer:
Cool, does it build okay locally? Have you tried triggering a build on travis?

PR #692, opened 2019-08-15
more robust BOOST_TOOLSET name guessing from from CC, CXX

lightmare:
There are two issues with `basename $CC`.

If `CC` is two words (e.g. `ccache gcc`), basename returns the first (the second word is treated as a suffix to remove, most often it doesn't match so you don't notice).
If `CC` is more than two words, basename complains there's too many arguments.

This PR first strips `ccache` if it's the first word, and then calls basename with only the first word of the rest.


springmeyer:
Thanks @lightmare!

PR #724, opened 2020-07-15
Add libpostal in mason

manoharuss:


manoharuss:
Hi @springmeyer when you have time, could you review this PR for installing libpostal from mason. 

build works for me locally on osx with
```
./mason build libpostal 1.1-alpha
```

I tried but I wasn't able to use `./mason trigger libpostal 1.1-alpha` as I did not have MASON_TRAVIS_TOKEN variable. Not sure about the hash mismatch on travis for expat.
https://travis-ci.org/github/mapbox/mason/builds/711237110

mattficke:
> Travis: The expat installation seems return a md5 sha mismatch. I do not completely understand what happens if i directly update the sha there. could you also help with that.

@manoharuss a mismatched hash like that usually means there was a problem with the download. I retried the build and it succeeded.

>I wasn't able to use ./mason trigger libpostal 1.1-alpha as I didn't not have MASON_TRAVIS_TOKEN variable.

You can use the [travis gem](https://github.com/travis-ci/travis.rb) to fetch this token, there are instructions in the readme: https://github.com/mapbox/mason#trigger

PR #734, opened 2020-12-01
Added libwebp and aws-sdk

corporategoth:


jakepruitt:
@corporategoth it looks like this failed to build due to timeout - should we move this out of "ready for review"?

Also I would love to see an explanation of why we need to switch the boost compile and any implications this may have on other boost-using systems.

PR #735, opened 2020-12-21
Patch property tree to work with no exceptions

averkhaturau:
Apply https://github.com/boostorg/property_tree/pull/20 and https://github.com/boostorg/property_tree/pull/21 to be able to use property tree without exceptions and rtti support


springmeyer:
We should never change/update an existing release. So the two options here are to create a new release, and call it `1.74.0-1` or just apply this to a new 1.75.0 release (since we need to package that anyway). I've done the latter at https://github.com/mapbox/mason/pull/736. So how about we go forward with https://github.com/mapbox/mason/pull/736?. Re-open this PR if for some reason you need 1.74.0 and can't adapt to using 1.75.0

PR #736, opened 2020-12-21
Boost 1.75.0

springmeyer:
This builds off https://github.com/mapbox/mason/pull/735 by:

 - carrying over the ptree patch to 1.75.0 from @averkhaturau 
 - removing the asio patch from @brunoabinader that is already in 1.75.0

I've tested the @averkhaturau patch locally by doing:

```
./mason build boost 1.75.0
```

And then viewing `mason_packages/.build/boost_1_75_0/boost/property_tree/detail/ptree_implementation.hpp` to confirm it applied okay.

This is ready to test by downstream users. @artemp I think depends on ptree, so would be good to confirm this does not break things for mapnik.

### DIFF
--- a/scripts/boost/1.74.0/patch.diff
+++ b/scripts/boost/1.74.0/patch.diff
@@ -63,3 +63,375 @@ index 9c7018c..bb073f4 100644
  
    template <typename Ex>
    static bool equal_ex(const any_executor_base& ex1,
+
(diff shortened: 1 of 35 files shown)


averkhaturau:
nav-native works with patched boost 1.75 fine. Thanks a lot!

artemp:
https://travis-ci.com/github/mapbox/mason/builds/210428436 :green_circle: :+1:

springmeyer:
Thanks everyone. Will merge this now.

